PR #1, opened 2025-02-20
chore: initialize project

c12i:
Closes https://github.com/holochain/holochain/issues/4712


.github/workflows/general.yaml
```diff
@@ -0,0 +1,59 @@
+name: "general"
```
ThetaSinner, line 1:
```suggestion
name: "test"
```

Or build, or something like that please - "general" doesn't tell me what this is for in the UI

.github/workflows/general.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
ThetaSinner, line 7:
```suggestion
  pull_request: {}
```

We want to build pull requests to any branch ideally

.github/workflows/general.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  cargo_test:
+    name: cargo_test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.84.0
```
ThetaSinner, line 21: These machines have `rustup`. Might be simpler to just provide a `rust-toolchain.toml` and let `rustup` grab the version rather than using an action where we have to keep in sync with what we're using locally

.github/workflows/general.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo test --no-fail-fast
+
+  rustfmt:
```
ThetaSinner, line 25: The time taken to install Rust and rebuild dependencies is going to outweigh the speed gain of running these as separate steps. I suggest that the workflow run these checks as steps rather than jobs

Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "holochain_http_gateway"
+version = "0.1.0-alpha.0"
+description = "The Holochain HTTP Gateway for providing a way to bridge from the web2 world into Holochain"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+homepage = "https://github.com/holochain/hc-http-gw"
+license = "Apache-2.0"
+repository = "https://github.com/holochain/hc-http-gw"
+readme = "./README.md"
```
ThetaSinner, line 9:
```suggestion
```

This is the default

Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+path = "./src/bin/hc-http-gw.rs"
+
+[dependencies]
+anyhow = "1.0.96"
+axum = "0.8.1"
+clap = { version = "4.5.30", features = ["derive", "env"] }
+holochain_client = "0.6.0"
+serde = { version = "=1.0.203", features = ["derive"] }
+thiserror = "2.0.11"
+tokio = { version = "1.43.0", features = ["full"] }
```
ThetaSinner, line 23: Prefer being selective about what features you need from tokio, it's a big project

README.md
```diff
@@ -1,2 +1,14 @@
...
+## License
+
+[![License: CAL 1.0](https://img.shields.io/badge/License-CAL%201.0-blue.svg)](https://github.com/holochain/cryptographic-autonomy-license)
+
+Copyright (C) 2020-2025, Holochain Foundation
+
+This program is free software: you can redistribute it and/or modify it under the terms of the license
+provided in the LICENSE file (CAL-1.0). This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE.
```
ThetaSinner, line 14:
```suggestion
```

Not CAL licensed. The gateway has to at least ephemerally handle crypto keys for signing but it I believe it does not hold identification keys in the sense that the CAL is designed to cover.

flake.nix
```diff
@@ -0,0 +1,48 @@
...
+      perSystem = { inputs', config, system, pkgs, lib, ... }:
+        {
+          formatter = pkgs.nixpkgs-fmt;
+
+          devShells.default = pkgs.mkShell {
+            packages = [
+              (with inputs'.holonix.packages; [
+                holochain
+                lair-keystore
+                rust
```
ThetaSinner, line 32: I see, this is going to conflict with a `rust-toolchain.toml` then. I'm not that keen on mixing these. If we're using Nix locally then we should be using it on CI too. I don't think we actually need this though. What we need is sweettest and I think we can entirely do without a flake

src/routes/healthz.rs
```diff
@@ -0,0 +1,4 @@
+#[tracing::instrument]
+pub async fn healthz() -> &'static str {
```
ThetaSinner, line 2: Why healthz?

src/routes/mod.rs
ThetaSinner, line 1: Prefer avoiding `mod.rs` and use a `routes.rs` one level up

src/service.rs
```diff
@@ -0,0 +1,52 @@
...
+        HcHttpGatewayService { router, address }
+    }
+
+    /// Get the socket address the service is configured to use
+    pub fn address(&self) -> SocketAddr {
+        self.address
+    }
+
+    /// Start the HTTP server and run until terminated
+    pub async fn run(self) -> anyhow::Result<()> {
```
ThetaSinner, line 39: Avoid using `anyhow` in library code. It's convenient but really hard to use with other error handling libraries in upstream code. You've set up `thiserror` already. Recommend defining a result type that uses the error for it and use that here instead

src/service.rs
```diff
@@ -0,0 +1,52 @@
...
+        initialize_tracing_subscriber("info");
```
ThetaSinner, line 40: This will panic if `run` is called more than once. It's probably not appropriate to do that in library code. Suggest moving this to the `main` and initializing once on startup. Let users of the library decide how they want to configure tracing

Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+tracing = "0.1.41"
+tracing-bunyan-formatter = "0.3.10"
+tracing-subscriber = { version = "0.3.19", features = [
+  "env-filter",
+  "json",
+  "registry",
+] }
+
+[dev-dependencies]
+reqwest = "0.12"
```
ThetaSinner, line 33: Not yet used?

When this is introduced, learning from the problems we've just had in Holochain, I'd like to configure it without default features and with `rustls`, so we don't end up needing openssl as a system dep

src/routes/healthz.rs
c12i: Just a convention I have adopted from previous use, largely inspired by kubernetes

https://kubernetes.io/docs/reference/using-api/health-checks/

Changed for brevity though

Cargo.toml
c12i: Yes, will be used extensively for tests.

> I'd like to configure it without default features and with rustls

Thanks for the heads up

flake.nix
c12i: I see, wasn't sure as well what the nix flake would be doing if we are not exporting a nix package now

.github/workflows/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  checks:
+    name: tests, fmt and clippy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
```
ThetaSinner, line 18: Is this still needed? `rustup` should be provided by the runner

.github/workflows/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+    branches: [main]
+  pull_request: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  checks:
+    name: tests, fmt and clippy
```
ThetaSinner, line 14: Minor but prefer putting these in the order the steps run

```suggestion
    name: fmt, clippy and tests
```

Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+tracing = "0.1.41"
+tracing-bunyan-formatter = "0.3.10"
+tracing-subscriber = { version = "0.3.19", features = [
+  "env-filter",
+  "json",
+  "registry",
+] }
+
+[dev-dependencies]
+reqwest = { version = "0.12", features = ["rustls-tls"] }
```
ThetaSinner, line 32: I think default features has to be disabled to get rid of the `openssl-sys` dependency but that can be figured out when we put this in Nix if that's fiddly to find now

ThetaSinner (approved):
Nice, looks good!

Issue #8, opened 2025-02-13
[httpgw] Accept HTTP GET requests


ThetaSinner:
Add a GET handler to the webserver and validate the incoming URL according to the rules in the spec.

```
http://<host>/<dna-hash>/<coordinator-idenfitier>/<zome-name>/<function-name>?payload=<payload>
```

Send back an OK Response for now

AC:
- There are tests for valid and invalid URLs
- Handle validation of URL params
- Ensure payload size is within HC_GW_PAYLOAD_LIMIT

PR #14, opened 2025-03-04
feat: accept http get requests

c12i:
This PR adds the initial GET handler for the zome call endpoint

Closes https://github.com/holochain/hc-http-gw/issues/8


src/routes/zome_call.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct PayloadQuery {
+    pub payload: Option<String>,
+}
+
+#[tracing::instrument]
+pub async fn zome_call(
+    Path(params): Path<ZomeCallParams>,
+    Query(query): Query<PayloadQuery>,
```
c12i, line 21: Axum handles the deserialization of the query params and query strings under the hood, handling URL encoded strings as well.

src/config.rs
```diff
@@ -14,15 +14,15 @@ use url2::Url2;
 use crate::{HcHttpGatewayError, HcHttpGatewayResult};
 
 /// Default payload size limit (10 kilobytes)
-pub const DEFAULT_PAYLOAD_LIMIT_BYTES: u32 = 10 * 1024;
+pub const DEFAULT_PAYLOAD_LIMIT_BYTES: usize = 10 * 1024;
```
c12i, line 17: Changed the type to `usize` for platform compatibility

Cargo.toml
```diff
@@ -30,4 +31,7 @@ tracing-subscriber = { version = "0.3.19", features = [
 url2 = "0.0.6"
 
 [dev-dependencies]
+holochain = { version = "0.4.1", features = ["test_utils"] }
```
c12i, line 34: Added this to use `fixt` in tests

src/routes/zome_call.rs
ThetaSinner: Actually same for the payload. It's not clear from the ticket whether checking that it decodes to JSON should be a requirement for this issue but it's not on a later ticket and it's a validation check. Should we be doing that now?

ThetaSinner: We should definitely be checking it can decode from base64

tests/integration/setup.rs
```diff
@@ -0,0 +1,57 @@
...
+            portpicker::pick_unused_port().unwrap(),
+            config.clone(),
+        );
+
+        let address = service.address().to_string();
+
+        // Run service in the background
+        tokio::task::spawn(async move { service.run().await.unwrap() });
+        // Wait for service to start
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
ThetaSinner, line 49: Avoid sleeping in tests as much as possible. Axum provide examples of how to deal with this kind of thing. They have some slick testing patterns but if we want to do this kind of integration testing then they demonstrate it by binding the `TcpListener` which means your client will always be able to connect immediately and the webserver will handle the request when it's ready to process messages off the socket. Entirely solid logic, shouldn't be flaky.

https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs#L130

tests/integration/zome_call.rs
```diff
@@ -0,0 +1,87 @@
...
+    config::{AllowedFns, Configuration},
+    tracing::initialize_tracing_subscriber,
+};
+use reqwest::StatusCode;
+
+use crate::TestApp;
+
+#[tokio::test]
+async fn zome_call_uses_correct_route_parameters() {
+    initialize_tracing_subscriber("info");
```
ThetaSinner, line 14: It's fine to have a default level for the tracing subscriber but as the person running the test, I'd like to be able to change the log level without changing the code

tests/integration/zome_call.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let app = TestApp::spawn().await;
+
+    let dna_hash = fixt!(DnaHash);
```
ThetaSinner, line 18: `fixt!` is okay, and it can be helpful for generating test data. I think for what we're doing here, it's probably not necessary to have a dependency on it. A DNA hash is just a byte array as far as we're concerned here. Doing `DnaHash::from_raw_36(vec![12; 36]);` will get the job done and be recognized as a `DnaHash` by the receiver.

When it comes to testing with a fixture hApp, we'll have a real DNA hash to use.

tests/integration/zome_call.rs
```diff
@@ -0,0 +1,87 @@
...
+        ))
+        .send()
+        .await
+        .expect("Failed to execute request");
+
+    assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+}
+
+#[tokio::test]
+async fn zome_call_with_small_payload_works() {
```
ThetaSinner, line 66: This kind of test where we're not needing a full integration setup might be a good place to look at Axum's other testing options. We just want to run the middleware and parse a request. We don't need networking and a real client/server setup.

Especially as we need some more tests. At least one that checks for an invalid DNA hash. Then ideally one that sends a payload that isn't valid JSON, though that depends on a discussion above.

tests/integration/zome_call.rs
c12i: I tried `DnaHash::from_raw_36` but I was encountering this error: `Holo Hash checksum validation failed`. I don't experience the same error with `fixt`

tests/integration/setup.rs
c12i: Thanks, removed the timeout and tests are passing still, this was really an unnecessary precautionary measure

tests/integration/zome_call.rs
c12i: Removed the param entirely, will revert to default if `RUST_LOG` is not set in the environment

tests/integration/setup.rs
```diff
@@ -0,0 +1,57 @@
...
+        .unwrap();
+
+        TestApp::spawn_with_config(config).await
+    }
+
+    /// Create a test app with custom configuration
+    pub async fn spawn_with_config(config: Configuration) -> Self {
+        let service = HcHttpGatewayService::new(
+            [127, 0, 0, 1],
+            portpicker::pick_unused_port().unwrap(),
```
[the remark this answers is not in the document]
c12i: Removed, using port `0` in tests

tests/integration/main.rs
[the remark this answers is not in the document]
c12i: Changed, and thanks for the tip, my style of organization was influenced by the book, [Zero to Production in Rust](https://www.zero2prod.com/index.html?country=Kenya&discount_code=AF60), good read btw

src/routes/zome_call.rs
```diff
@@ -0,0 +1,65 @@
...
+        let dna_hash = DnaHash::try_from(raw_params.dna_hash)?;
+
+        Ok(ZomeCallParams {
+            dna_hash,
+            coordinator_identifier: raw_params.coordinator_identifier,
+            zome_name: raw_params.zome_name,
+            function_name: raw_params.function_name,
+        })
+    }
+}
```
c12i, line 45: The default extractor was not producing invalid length errors when trying to convert the dna hash query param to a `DnaHash` type. Invalid DNA hashes now produce `400` error responses

https://github.com/holochain/hc-http-gw/pull/14#discussion_r1981349232

tests/integration/zome_call.rs
c12i: Add some extra tests for checking for valid json from the decoded payload, and a test to verify that we can derive the `DnaHash` from the URL params

tests/integration/setup.rs
```diff
@@ -0,0 +1,57 @@
...
+            config.clone(),
+        );
+
+        let address = service.address().to_string();
+
+        // Run service in the background
+        tokio::task::spawn(async move { service.run().await.unwrap() });
```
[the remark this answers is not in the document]
c12i: Thanks for the tip, fixed

src/service.rs
```diff
@@ -20,17 +22,28 @@ pub struct HcHttpGatewayService {
...
+        let state = AppState { configuration };
+
+        let zome_call_routes = Router::new()
+            .route(
+                "/{dna_hash}/{coordinator_identifier}/{zome_name}/{function_name}",
+                get(zome_call),
+            )
+            .layer(middleware::from_fn_with_state(
+                state.clone(),
+                validate_zome_call_payload,
```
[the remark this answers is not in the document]
c12i: Renamed

src/config.rs
c12i: > Does it need to ever be a u64?

Not really, using usize seems like the more the idiomatic Rust choice for memory sizes and counts as it automatically adjusts to the platform

ThetaSinner: It's more important that the code accurately expresses intent. If it's never valid for this to be larger than `u32::MAX` then I think we're losing clarity by choosing `usize`

src/error.rs
```diff
@@ -9,7 +16,59 @@ pub enum HcHttpGatewayError {
...
+                tracing::error!("Path deserialization error: {}", e);
+                error_from_status(400, Some("Invalid Request Path"))
+            }
+            HcHttpGatewayError::Base64DecodingError(e) => {
+                tracing::error!("Base64 decode error: {}", e);
+                error_from_status(400, Some("Failed to decode base64 encoded strig"))
+            }
+            HcHttpGatewayError::HoloHashError(e) => {
+                tracing::error!("HoloHash error: {}", e);
+                error_from_status(400, Some("HoloHash error"))
```
ThetaSinner, line 49: At the moment this can only be "Invalid base64 DNA hash", I'd rather have that feedback as a user. I think it might make sense to inline some of this error handling to keep the error messages specific where appropriate

src/middleware/validate_zome_call_payload_size.rs
```diff
@@ -0,0 +1,53 @@
...
+    http::StatusCode,
+    middleware::Next,
+    response::{IntoResponse, Response},
+    Json,
+};
+
+use crate::{http::HcGwErrorResponse, routes::PayloadQuery, service::AppState};
+
+#[tracing::instrument(skip(state, request, next))]
+pub async fn validate_zome_call_payload_size(
```
ThetaSinner, line 12: I'm still a little torn about this as middleware. Middleware is supposed to be re-usable. So checking authentication or some other request headers, or pre-processing request content, or handling upgrading requests etc. 

At the moment we don't have requirements for any other requests that need to check a payload size. Even if we do, this is tied to a configuration setting so it's not necessarily re-usable.

I'm not asking for this to be changed right now, but I think we could save some complexity by merging it into the GET request handler.

src/routes/zome_call.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct PayloadQuery {
+    pub payload: Option<String>,
+}
+
+#[tracing::instrument]
+pub async fn zome_call(
```
ThetaSinner, line 53: Nice, yes this is a solid check

src/service.rs
```diff
@@ -1,56 +1,74 @@
...
-        HcHttpGatewayService { router, address }
+        let listener = TcpListener::bind(address).await?;
+
+        Ok(HcHttpGatewayService { router, listener })
     }
 
     /// Get the socket address the service is configured to use
-    pub fn address(&self) -> SocketAddr {
-        self.address
+    pub fn address(&self) -> HcHttpGatewayResult<SocketAddr> {
```
ThetaSinner, line 60: Nice, this keeps the listener encapsulated and achieves the same result as the axum testing example. I like it

tests/setup/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        let service = HcHttpGatewayService::new([127, 0, 0, 1], 0, config.clone())
+            .await
+            .unwrap();
+
+        let address = service.address().unwrap().to_string();
+
+        // Run service in the background
+        let task_handle = tokio::task::spawn(async move { service.run().await.unwrap() });
+        // Wait for service to start
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
ThetaSinner, line 48: Is this still required? If we're able to get the local address above then we should be able to send TCP traffic to the service even if it hasn't fully started yet? Anything we send should get buffered and Axum will read it once it starts.

tests/integration/zome_call.rs
ThetaSinner: I wasn't clear enough there. I linked to a specific test as an example of making the test non-flaky without a sleep. That same file contains a bunch of other tests though. If you've looked at those and decided that there's no value there then that's fine, I just want to make sure I've clearly communicated that there are some other examples of just testing a router without booting up a whole server/client.

tests/integration/zome_call.rs
ThetaSinner: Oh that's interesting. We have many tests where we do that but when we deserialize from string, the location bytes get checked. In that case, you can still use `from_raw_32`, that will calculate valid location bytes from the input.

tests/integration/setup.rs
ThetaSinner: Commented separately, but this still shows up for me. Maybe a local change not pushed yet?

src/routes/zome_call.rs
ThetaSinner: All my comments on this thread are resolved, I'll leave it open because I didn't start it

src/error.rs
```diff
@@ -9,7 +16,59 @@ pub enum HcHttpGatewayError {
     /// Handles configuration parsing errors
     #[error("Configuration error: {0}")]
     ConfigurationError(String),
+    /// Handle path deserialization errors
+    #[error("Path deserialization error: {0}")]
+    PathDeserializationError(#[from] axum::extract::rejection::PathRejection),
```
jost-s, line 21: I'd call this something like `PathResolutionError` or `PathParsingError`.

tests/setup/mod.rs
jost-s, line 48: Instead of a fixed sleep, it's more reliable to set up a retry loop with a timeout, e.g.
```rust
tokio::time::timeout(Duration::from_millis(100), async move {
  loop {
    tokio::time::sleep(Duration::from_millis(10).await;
    if call_health_endpoint().await.is_ok() {
      break;
    }
  }
}
```

tests/zome_call.rs
```diff
@@ -0,0 +1,126 @@
...
+        ))
+        .send()
+        .await
+        .expect("Failed to execute request");
+
+    assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+}
+
+#[tokio::test]
+async fn zome_call_with_invalid_json_payload() {
```
jost-s, line 73: There should be a happy path test too with a valid payload.

jost-s (changes requested):
Looks pretty good to me. I request to add a test with a valid payload.

tests/setup/mod.rs
c12i: Oops, not required anymore, removed

c12i: I have removed the requirement for sleeping, so this is not needed anymore

tests/zome_call.rs
c12i: Added

src/middleware/validate_zome_call_payload_size.rs
c12i: I see, I have moved the payload checks to the route for simplicity

src/config.rs
c12i: I see, reverted to `u32`

tests/integration/setup.rs
c12i: Yes this stuck around while moving around the modules, fixed now

src/error.rs
c12i: Renamed to the latter

src/error.rs
c12i: Updated the error message.

> I think it might make sense to inline some of this error handling to keep the error messages specific where appropriate

Not sure what you mean with this, are you referring to not using the `error_from_status` function building the axum error response for each error variant?

tests/setup/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        TestApp {
+            address,
+            client: Client::new(),
+            task_handle,
+        }
+    }
+
+    /// Util to make a request to the zome call GET endpoint
+    #[allow(unused, reason = "Actually used, only suppressing warning")]
+    pub async fn call_zome(
```
c12i, line 57: Added this function to keep test code DRY

tests/integration/zome_call.rs
ThetaSinner: What is the outcome here, are we sticking with the current testing pattern?

tests/integration/zome_call.rs
ThetaSinner: This thread is still open

tests/integration/zome_call.rs
c12i: Yea, I am fine to stick with it, especially after the refactor to how the TCP listern is handled

tests/integration/zome_call.rs
c12i: Oh my bad, I tried creating a DnaHash with `from_raw_32` in the tests too, but this yielded a panic from assertion `failed: invalid byte count for HoloHash [219, 32]`

src/error.rs
c12i: > I'm wondering if we should handle the DnaHash error inline rather than converting it to an error type that doesn't have context and then to an error message

I see. I think this would make sense if we were decoding Dna hashes in multiple places in the server, since for now we are doing it in one place, which I am not sure we will be doing, hence why I moved the error handling to the error module.

src/error.rs
```diff
@@ -9,7 +16,80 @@ pub enum HcHttpGatewayError {
     /// Handles configuration parsing errors
     #[error("Configuration error: {0}")]
     ConfigurationError(String),
+    /// Handle path deserialization errors
+    #[error("Path deserialization error: {0}")]
```
jost-s, line 20:
```suggestion
    /// Handle path parsing errors
    #[error("Path parsing error: {0}")]
```

jost-s (approved):
Great!

tests/integration/zome_call.rs
ThetaSinner: That's odd. I'd have to give that a try myself to understand why. This definitely should be possible to construct without having a valid keypair.

ThetaSinner (approved):

PR #15, opened 2025-03-05
feat: establish admin ws connection

### LINKED ISSUE #11: [httpgw] Establish admin ws connection
AC:

- On startup establish admin ws connection
- Connection and reconnection logic implemented for the admin websocket
- Tests: connection and reconnection

c12i:
Closes #11


src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    sync::{Arc, Mutex},
+};
+
+use holochain_client::{AdminWebsocket, ConductorApiError, ConductorApiResult};
+use tokio::time::{sleep, Duration};
+
+use crate::{HcHttpGatewayError, HcHttpGatewayResult};
+
+const ADMIN_WS_CONNECTION_MAX_RETRIES: usize = 1;
+const ADMIN_WS_CONNECTION_RETRY_DELAY_MS: u64 = 1000;
```
c12i, line 12: Not sure if this is necessary, but curious to know if it would be required

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// A wrapper around AdminWebsocket that automatically handles reconnection
+/// when the connection is lost due to network issues or other failures.
+#[derive(Clone)]
+pub struct ReconnectingAdminWebsocket {
+    /// The WebSocket URL to connect to
+    url: String,
```
ThetaSinner, line 19: String typed? Should this either be a URL or a `(IpAddr, u16)`?

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// The handle to the AdminWebsocket connection
+    connection_handle: Arc<Mutex<Option<AdminWebsocket>>>,
+    /// Current retry attempt counter
+    current_retries: usize,
```
ThetaSinner, line 23: Unclear field on a `Clone` type that claims to be thread-safe. Is this really safe to clone or should it be an `Arc<AtomicXXX>`?

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+            connection_handle: Arc::new(Mutex::new(None)),
+            current_retries: 0,
+        }
+    }
+
+    /// Checks if there is an active connection
+    fn is_connected(&self) -> HcHttpGatewayResult<bool> {
+        let connection = self.connection_handle.lock().map_err(|e| {
+            HcHttpGatewayError::InternalError(format!(
+                "Mutex was poisoned during is_connected check: {}",
```
ThetaSinner, line 40: This only happens if there has been a panic elsewhere while the lock was held and it's not recoverable. It's letting you handle the error in case you want to replace the mutex or otherwise resolve the problem state yourself but bubbling the state up is not solving the issue. I'd recommend using `.expect("some message")` here 

https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning

tests/websocket.rs
```diff
@@ -23,3 +25,37 @@ async fn connect_app_websocket() {
...
+
+    let apps = admin_ws
+        .call(|ws| async move { ws.list_apps(None).await })
+        .await;
+
+    if let Err(err) = apps {
+        assert!(matches!(err, HcHttpGatewayError::UpstreamUnavailable));
+    } else {
+        panic!("Expected UpstreamUnavailable error, found: {apps:?}");
+    }
```
jost-s, line 60: Please also test that the admin websocket can reconnect successfully after this error.

src/admin_ws.rs
jost-s: I don't think there has to be a delay before reconnecting. If there is a reason, it's not possible to know how long the delay should be.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+        self.current_retries
+    }
+
+    /// Ensures that a connection is established before proceeding
+    async fn ensure_connected(&mut self) -> HcHttpGatewayResult<()> {
+        if self.is_connected()? {
+            return Ok(());
+        }
+
+        self.connect().await
```
ThetaSinner, line 59: We had to take out a lock in `is_connected`, then another one to open the connection and store it in the mutex, and then we return which means we have to take out a third lock to use the created object. There is a potential for race conditions to do it this way, and taking out locks should be considered expensive.

This should be possible within a single lock:
1. Take out a lock
2. If there is a connection, clone and return it
3. If there is no connection, establish one and store a clone, then return the original

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use crate::{HcHttpGatewayError, HcHttpGatewayResult};
+
+const ADMIN_WS_CONNECTION_MAX_RETRIES: usize = 1;
+const ADMIN_WS_CONNECTION_RETRY_DELAY_MS: u64 = 1000;
+
+/// A wrapper around AdminWebsocket that automatically handles reconnection
+/// when the connection is lost due to network issues or other failures.
+#[derive(Clone)]
+pub struct ReconnectingAdminWebsocket {
```
jost-s, line 17: The "reconnecting" is more an implementation detail. If you want to use a special denominator, I suggest
```suggestion
pub struct HcHttpGwAdminWebsocket {
```

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    // Helper method to execute an operation with the current connection
+    async fn exec<T, F, Fut>(&mut self, f: &F) -> HcHttpGatewayResult<T>
+    where
+        F: Fn(Arc<AdminWebsocket>) -> Fut + Send,
+        Fut: Future<Output = ConductorApiResult<T>> + Send,
+    {
+        // block scope to limit MutexGuard lifetime
+        let connection = {
+            let mut connection = self.connection_handle.lock().map_err(|e| {
```
jost-s, line 134: Mutexes can be unwrapped. It's assumed that it's not been poisoned by a panic. If there's a way to panic, I think we better eliminate that than treating this error.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Attempts to connect to the AdminWebsocket
+    ///
+    /// This will make multiple attempts according to the `max_retries` and
+    /// `retry_delay_ms` settings, with exponential backoff between retries.
+    pub async fn connect(&mut self) -> HcHttpGatewayResult<()> {
+        self.current_retries = 0;
+
+        while self.current_retries < ADMIN_WS_CONNECTION_MAX_RETRIES {
+            match AdminWebsocket::connect(&self.url).await {
```
ThetaSinner, line 70: This function is re-entrant and I don't think it should be. The point of this type is to maintain a single admin connection but we're not protecting this operation by locking. 

We only take out a lock if the connection attempt succeeds which means we might connect multiple clients and then assign one and close the others.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+                    self.current_retries += 1;
+                    tracing::warn!(
+                        "Failed to connect to WebSocket (attempt {}/{}): {:?}",
+                        self.current_retries,
+                        ADMIN_WS_CONNECTION_MAX_RETRIES,
+                        e
+                    );
+
+                    if self.current_retries < ADMIN_WS_CONNECTION_MAX_RETRIES {
+                        sleep(Duration::from_millis(ADMIN_WS_CONNECTION_RETRY_DELAY_MS)).await;
```
ThetaSinner, line 94: Why delay between connection attempts? There is an HTTP request waiting for this to happen. We should be making one attempt to connect if we ran into an error and otherwise we just return an error and free up resources to process a new request.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Attempts to connect to the AdminWebsocket
+    ///
+    /// This will make multiple attempts according to the `max_retries` and
+    /// `retry_delay_ms` settings, with exponential backoff between retries.
+    pub async fn connect(&mut self) -> HcHttpGatewayResult<()> {
+        self.current_retries = 0;
+
+        while self.current_retries < ADMIN_WS_CONNECTION_MAX_RETRIES {
```
ThetaSinner, line 69: Loop only actually runs once, doesn't need to be a loop?

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+            ADMIN_WS_CONNECTION_MAX_RETRIES
+        )))
+    }
+
+    /// Allows calling a method on the AdminWebsocket, with automatic reconnection if needed
+    ///
+    /// Accepts a function `f` that takes a AdminWebsocket and returns a Result with ConductorApiError
+    pub async fn call<T, F, Fut>(&mut self, f: F) -> HcHttpGatewayResult<T>
+    where
+        F: Fn(Arc<AdminWebsocket>) -> Fut + Send + Clone,
```
ThetaSinner, line 113: `AdminWebsocket` is `Clone`, why put it behind an `Arc`?

ThetaSinner, line 113: Any reason not to use a `BoxFuture` here? Deals with getting the trait bounds right so it's simpler to work with

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+        Err(HcHttpGatewayError::ConfigurationError(format!(
+            "Maximum connection retry attempts ({}) reached",
+            ADMIN_WS_CONNECTION_MAX_RETRIES
+        )))
+    }
+
+    /// Allows calling a method on the AdminWebsocket, with automatic reconnection if needed
+    ///
+    /// Accepts a function `f` that takes a AdminWebsocket and returns a Result with ConductorApiError
+    pub async fn call<T, F, Fut>(&mut self, f: F) -> HcHttpGatewayResult<T>
```
ThetaSinner, line 111: I think this only needs one generic parameter, and that will make using this behind a trait slightly easier

```
pub async fn call<T>(&self, f: fn(AdminWebsocket) -> BoxFuture<'static, HcHttpGatewayResult<T>>;
```

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+        F: Fn(Arc<AdminWebsocket>) -> Fut + Send,
+        Fut: Future<Output = ConductorApiResult<T>> + Send,
+    {
+        // block scope to limit MutexGuard lifetime
+        let connection = {
+            let mut connection = self.connection_handle.lock().map_err(|e| {
+                HcHttpGatewayError::InternalError(format!("Connection mutex was poisoned: {e}"))
+            })?;
+
+            let connection = connection.take().ok_or_else(|| {
```
ThetaSinner, line 138: `.take()`? Doesn't that mean we have to open a new connection for every request?

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+pub struct ReconnectingAdminWebsocket {
+    /// The WebSocket URL to connect to
+    url: String,
+    /// The handle to the AdminWebsocket connection
+    connection_handle: Arc<Mutex<Option<AdminWebsocket>>>,
+    /// Current retry attempt counter
+    current_retries: usize,
+}
+
+impl ReconnectingAdminWebsocket {
```
jost-s, line 26: I think all this admin ws should expose is `connect` and `call`. `connect` could be an async constructor that immediately connects to the Admin API. No need for an `Option<AdminWebsocket>` then. If no connection to the Admin API can be established, there's no point to start up the service and the process can shut down.

Once it's connected, only `call` is needed to list apps, issue tokens, create signing credentials and create the app websocket connections. How many retries there have been and how reconnections are handled is not of concern to the consumer of this struct.

When `call` fails because of a disconnect, it reconnects and calls again. If the call fails repeatedly, it means the conductor cannot be reached anymore. That's probably a case where the gateway (and the conductor) have to be restarted.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl HcHttpGatewayErrorExt for HcHttpGatewayError {
+    fn is_disconnect_error(&self) -> bool {
+        match self {
+            // Specifically checking for WebsocketError (inside ConductorApiError) and IoError as mentioned
+            HcHttpGatewayError::ConductorApiError(api_error) => {
+                // Check for websocket errors inside the ConductorApiError
+                matches!(api_error, ConductorApiError::WebsocketError(_))
+            }
+            HcHttpGatewayError::IoError(_) => true,
```
ThetaSinner, line 182: How do we get I/O errors that don't come back through a `WebsocketError`?

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    future::Future,
+    sync::{Arc, Mutex},
+};
+
+use holochain_client::{AdminWebsocket, ConductorApiError, ConductorApiResult};
+use tokio::time::{sleep, Duration};
+
+use crate::{HcHttpGatewayError, HcHttpGatewayResult};
+
+const ADMIN_WS_CONNECTION_MAX_RETRIES: usize = 1;
```
jost-s, line 11: 1 is far too small. That means that as soon as there's been one error, the ws is reconnected and that's it. Another error and the gateway is out of service. I don't think the retries need to be tracked at all. Instead there can be 2-3 reconnection attempts and if they have failed, the gateway could shut down.

src/error.rs
```diff
@@ -51,6 +51,16 @@ pub enum HcHttpGatewayError {
         /// Function name
         fn_name: String,
     },
+
+    /// Handles conductor api errors
+    #[error("Conductor API error: {0}")]
+    ConductorApiError(#[from] holochain_client::ConductorApiError),
+    /// Handles miscellaneous internal errors
+    #[error("Internal Error: {0}")]
+    InternalError(String),
```
ThetaSinner, line 60: Is this still needed if we don't bubble poison errors?

jost-s (changes requested):
I think the admin websocket can be simplified and the exposed functions should be limited. How the websocket handles reconnections doesn't need to be known by the service. Without the admin websocket connection, the gateway is pretty much useless, so I don't see the need for it to keep running without it.

src/router.rs
```diff
@@ -2,11 +2,18 @@ use crate::{
     config::Configuration,
     routes::{health_check, zome_call},
     service::AppState,
+    ReconnectingAdminWebsocket,
 };
 use axum::{http::StatusCode, routing::get, Router};
 
 pub fn hc_http_gateway_router(configuration: Configuration) -> Router {
-    let state = AppState { configuration };
+    let admin_ws = ReconnectingAdminWebsocket::new(configuration.admin_ws_url.as_ref());
```
ThetaSinner, line 10: This will need to be a parameter to `hc_http_gateway_router` as a trait. See https://github.com/holochain/hc-http-gw/pull/23

Otherwise we can't do unit testing that doesn't connect to Holochain

src/service.rs
```diff
@@ -16,9 +16,11 @@ pub struct HcHttpGatewayService {
 }
 
 /// Shared application state
-#[derive(Debug, Clone)]
+#[derive(Clone)]
```
ThetaSinner, line 19: Why not permit debug?

tests/setup/mod.rs
```diff
@@ -1,10 +1,10 @@
+use std::collections::HashMap;
+
```
ThetaSinner, line 2:
```suggestion
```

tests/websocket.rs
```diff
@@ -23,3 +25,37 @@ async fn connect_app_websocket() {
...
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn connect_admin_websocket() {
+    let mut sweet_conductor = SweetConductor::from_standard_config().await;
+
+    let admin_port = sweet_conductor
+        .get_arbitrary_admin_websocket_port()
+        .unwrap();
+    let url = format!("localhost:{admin_port}");
```
ThetaSinner, line 36: Not a valid URL, doesn't have a scheme, this is why I'm asking for this to be something other than a string above.

tests/websocket.rs
```diff
@@ -23,3 +25,37 @@ async fn connect_app_websocket() {
...
+
+    let mut admin_ws = ReconnectingAdminWebsocket::new(&url);
+
+    // First call should succeed
+    let apps = admin_ws
+        .call(|ws| async move { ws.list_apps(None).await })
+        .await
+        .unwrap();
+
+    assert_eq!(apps.len(), 0);
```
ThetaSinner, line 46:
```suggestion
    assert!(apps.is_empty());
```

tests/websocket.rs
```diff
@@ -23,3 +25,37 @@ async fn connect_app_websocket() {
...
+    assert_eq!(admin_ws.get_reconnection_retries(), 0);
+
+    // Shutdown the conductor to force a connection error
+    sweet_conductor.shutdown().await;
+
+    let apps = admin_ws
```
ThetaSinner, line 52: We're not then testing that it successfully reconnects if Holochain is started back up, so we're missing the happy path test.

src/admin_ws.rs
ThetaSinner: I've left some similar comments about this retry logic

src/admin_ws.rs
ThetaSinner: Agree about limiting the API surface, and the trait proposed on #23 should help with that.

Then the `Option` is actually quite helpful I think because it means we can startup even if we can't get connected to Holochain. I'd actually like to keep that and establish the connection on demand rather than on startup. At least that's how things like to me after working on #23 and thinking about how this will get plumbed together in the binary.

ThetaSinner:
The test `connect_admin_websocket` doesn't pass for me locally, something I'm doing wrong?

```
thread 'connect_admin_websocket' panicked at tests/websocket.rs:66:18:
Expected WebsocketError but got: Err(ConfigurationError("Maximum connection retry attempts (1) reached"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/admin_ws.rs
jost-s: I see now that it's only used inside of `connect` and reset there at the beginning, so it doesn't need to be a field at all.

src/admin_ws.rs
```diff
@@ -0,0 +1,187 @@
...
+const ADMIN_WS_CONNECTION_RETRY_DELAY_MS: u64 = 1000;
+
+/// A wrapper around AdminWebsocket that automatically handles reconnection
+/// when the connection is lost due to network issues or other failures.
+#[derive(Clone)]
+pub struct ReconnectingAdminWebsocket {
+    /// The WebSocket URL to connect to
+    url: String,
+    /// The handle to the AdminWebsocket connection
+    connection_handle: Arc<Mutex<Option<AdminWebsocket>>>,
```
ThetaSinner, line 21: We should be using an `RwLock` here and cloning `AdminWebsocket` handles out from behind the lock I think. That means we can mostly use `.read()` without blocking the caller waiting for other requests to get an admin connection.

All this really is behind the scenes is a queue for pushing messages onto the network. It's fine to have lots of threads doing that at once.

src/admin_ws.rs
jost-s: Okay, got it, then only `new` to store the URL and `call` to handle connections and make calls are needed.

src/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum HcHttpGatewayError {
     /// Error returned when a connection cannot be made to the upstream Holochain service
     #[error("The upstream Holochain service could not be reached")]
     UpstreamUnavailable,
+    /// Handles miscellaneous internal errors
+    #[error("Internal Error: {0}")]
+    InternalError(String),
```
mattyg, line 34: I'd prefer this split into specific errors. It looks like its being thrown for two distinct cases: invalid configuration, and failure to connect to admin websocket (which maybe should throw the existing `UpstreamUnavailable` variant).

src/admin_ws.rs
```diff
@@ -0,0 +1,203 @@
...
+const ADMIN_WS_CONNECTION_MAX_RETRIES: usize = 1;
+
+/// A wrapper around AdminWebsocket that automatically handles reconnection
+/// when the connection is lost due to network issues or other failures.
+#[derive(Debug, Clone)]
+pub struct HcHttpGwAdminWebsocket {
+    /// The WebSocket URL to connect to
+    url: Url2,
+    /// The handle to the AdminWebsocket connection
+    connection_handle: Arc<RwLock<Option<AdminWebsocket>>>,
```
mattyg, line 17: Why is this an `Option<AdminWebsocket>`? It looks like it needs to make a successful connection within the `connect` constructor, so if the struct is instantiated then we should always have an `AdminWebsocket`

src/admin_ws.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// Attempts to connect to the AdminWebsocket
+    ///
+    /// This will make multiple attempts according to the `max_retries`
+    /// setting, with exponential backoff between retries.
+    async fn attempt_connection(&mut self) -> HcHttpGatewayResult<()> {
+        self.current_retries = 0;
+
+        while self.current_retries < ADMIN_WS_CONNECTION_MAX_RETRIES {
```
mattyg, line 65:
```suggestion
        while self.current_retries <= ADMIN_WS_CONNECTION_MAX_RETRIES {
```

I think this is the intention

ThetaSinner:
Updated by #27

PR #16, opened 2025-03-06
chore: add Nix devShell with direnv support

cdunster:


flake.nix
```diff
@@ -0,0 +1,42 @@
...
+        _module.args.pkgs = import inputs.nixpkgs {
+          inherit system;
+          overlays = [ inputs.rust-overlay.overlays.default ];
+        };
+
+        formatter = pkgs.nixpkgs-fmt;
+
+        devShells.default =
+          let
+            rustFromFile = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
```
cdunster, line 30: I used the Rust toolchain from the file because there already was one, but maybe I should use the one from Holonix?

flake.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  inputs = {
+    nixpkgs.url = "nixpkgs/nixos-unstable";
+
+    flake-parts.url = "github:hercules-ci/flake-parts";
+
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.flake-utils.follows = "flake-utils";
```
jost-s, line 12: There are no `flake-utils` here.

jost-s (dismissed):

flake.nix
ThetaSinner: I like this the way it is, no need for it to match Holonix here if we're not depending on Holonix.

Also, this binary might end up in Holonix, so it's nice to not have a circular dependency of sorts.

flake.nix
cdunster: oops, nice catch :smile: I used my template flake but then swapped `flake-utils` for `flake-parts` and it looks like I missed some.

ThetaSinner (approved):

PR #17, opened 2025-03-06
Add app websocket management

### LINKED ISSUE #10: [httpgw] Implement logic for app websocket management
Implement the logic for app websocket management. Given an app to connect to, select or add an appropriate app interface

AC:
- This logic should be similar or the same as what the [`hc` sandbox](https://github.com/holochain/holochain/blob/develop/crates/hc_sandbox/src/zome_call.rs#L361) does, just with a different origin
- There is a limit on how many open app websockets the gateway will maintain to Holochain
   - Set MAX_APP_ID_CONNECTION as configurable env var (500 limit for now)
   - We keep a handle of the connected app ws handles as a hash map keyed by the target app id
- If an app websocket is closed, it should be possible to open a new one rather than failing
   - Max retries: 1, an appropriate 50_ response on error
- Tests: app interface creation

ThetaSinner:
Closes #10

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,6 +28,8 @@ jobs:
         run: |
           nix develop -c ./fixture/package.sh
       - name: Run tests
+        env:
+          RUST_LOG: warn,holochain_http_gateway=info
         run: cargo test --no-fail-fast -- --test-threads=1
 
   ci_pass:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,7 +28,7 @@ jobs:
         run: |
           nix develop -c ./fixture/package.sh
       - name: Run tests
-        run: cargo test --no-fail-fast
+        run: cargo test --no-fail-fast -- --test-threads=1
 
   ci_pass:
     if: ${{ always() }}
--- a/src/app_conn_pool.rs
+++ b/src/app_conn_pool.rs
@@ -4,7 +4,7 @@ use futures::future::BoxFuture;
 use holochain_client::{
     AdminWebsocket, AppWebsocket, AuthorizeSigningCredentialsPayload, CellInfo, ClientAgentSigner,
     ConductorApiError, ConnectRequest, GrantedFunctions, IssueAppAuthenticationTokenPayload,
-    WebsocketConfig,
+    Timestamp, WebsocketConfig,
 };
 use holochain_types::app::InstalledAppId;
 use holochain_types::websocket::AllowedOrigins;
@@ -15,6 +15,16 @@ use std::sync::{Arc, RwLock};
 /// The origin that the gateway will use when connecting to Holochain app interfaces.
 pub const HTTP_GW_ORIGIN: &str = "hc-http-gw";
 
+/// A wrapper around an app websocket connection that includes state required to manage the
+/// connection.
+#[derive(Debug, Clone)]
+pub struct AppWebsocketWithState {
+    /// The app websocket connection.
+    pub app_ws: AppWebsocket,
+    /// The time at which the connection was opened.
+    pub opened_at: Timestamp,
+}
+
 /// A connection pool for app connections.
 ///
 /// This is a pool in the sense that it manages multiple connections to Holochain app interfaces,
@@ -23,7 +33,7 @@ pub const HTTP_GW_ORIGIN: &str = "hc-http-gw";
 pub struct AppConnPool {
     configuration: Configuration,
     cached_app_port: Arc<RwLock<Option<u16>>>,
-    app_clients: Arc<tokio::sync::RwLock<HashMap<InstalledAppId, AppWebsocket>>>,
+    app_clients: Arc<tokio::sync::RwLock<HashMap<InstalledAppId, AppWebsocketWithState>>>,
 }
 
 impl AppConnPool {
@@ -50,10 +60,10 @@ impl AppConnPool {
         execute: impl Fn(AppWebsocket) -> BoxFuture<'static, HcHttpGatewayResult<T>>,
     ) -> HcHttpGatewayResult<T> {
         for _ in 0..2 {
-            let app_client = self
+            let app_ws = self
                 .get_or_connect_app_client(installed_app_id.clone(), admin_websocket.clone())
                 .await?;
-            match execute(app_client).await {
+            match execute(app_ws).await {
                 Ok(response) => {
                     return Ok(response);
                 }
@@ -87,36 +97,49 @@ impl AppConnPool {
         {
             let app_clients = self.app_clients.read().await;
 
-            println!(
-                "Trying to get client for app {} from pool: {:?}",
-                installed_app_id, app_clients
-            );
-
             if let Some(client) = app_clients.get(&installed_app_id) {
-                return Ok(client.clone());
+                return Ok(client.app_ws.clone());
             }
         }
 
-        match self
-            .app_clients
-            .write()
-            .await
-            .entry(installed_app_id.clone())
-        {
+        let mut app_client_lock = self.app_clients.write().await;
+
+        let app_ws = match app_client_lock.entry(installed_app_id.clone()) {
             std::collections::hash_map::Entry::Occupied(client) => {
                 // Created by another thread while we were waiting for the lock
-                Ok(client.get().clone())
+                client.get().app_ws.clone()
             }
             std::collections::hash_map::Entry::Vacant(entry) => {
-                let app_client = self
-                    .attempt_connect_app_client(installed_app_id, admin_ws, 1)
+                let app_ws = self
+                    .attempt_connect_app_ws(installed_app_id, admin_ws, 1)
                     .await?;
 
-                entry.insert(app_client.clone());
+                entry.insert(AppWebsocketWithState {
+                    app_ws: app_ws.clone(),
+                    opened_at: Timestamp::now(),
+                });
 
-                Ok(app_client)
+                app_ws
             }
+        };
+
+        if app_client_lock.len() > self.configuration.max_app_connections as usize {
+            // Find and remove the oldest connection
+            let installed_app_id = app_client_lock
+                .iter()
+                .min_by_key(|(_, v)| v.opened_at)
+                .map(|(k, _)| k.clone())
+                .expect("Invalid lock");
+
+            tracing::warn!(
+                "Reached maximum app connections, removing connection for app: {}",
+                installed_app_id
+            );
+
+            app_client_lock.remove(&installed_app_id);
         }
+
+        Ok(app_ws)
     }
 
     /// Remove an app client from the pool.
@@ -124,7 +147,7 @@ impl AppConnPool {
         self.app_clients.write().await.remove(installed_app_id);
     }
 
-    async fn attempt_connect_app_client(
+    async fn attempt_connect_app_ws(
         &self,
         installed_app_id: InstalledAppId,
         admin_ws: AdminWebsocket,
@@ -175,7 +198,7 @@ impl AppConnPool {
         let client_signer = ClientAgentSigner::default();
 
         // Attempt to connect to the app websocket
-        let app_client = match AppWebsocket::connect_with_request_and_config(
+        let app_ws = match AppWebsocket::connect_with_request_and_config(
             request,
             Arc::new(config),
             issued.token,
@@ -192,7 +215,7 @@ impl AppConnPool {
                 *self.cached_app_port.write().expect("Invalid lock") = None;
 
                 // Try again, with one fewer retry permitted
-                return Box::pin(self.attempt_connect_app_client(
+                return Box::pin(self.attempt_connect_app_ws(
                     installed_app_id,
                     admin_ws,
                     retries - 1,
@@ -202,7 +225,7 @@ impl AppConnPool {
         };
         tracing::debug!("Connected to app websocket");
 
-        let app_info = app_client.cached_app_info();
+        let app_info = app_ws.cached_app_info();
         let cells = app_info
             .cell_info
             .values()
@@ -249,7 +272,7 @@ impl AppConnPool {
             client_signer.add_credentials(cell_id, credentials);
         }
 
-        Ok(app_client)
+        Ok(app_ws)
     }
 
     async fn get_app_port(
@@ -290,7 +313,7 @@ impl AppConnPool {
     #[cfg(feature = "test-utils")]
     pub fn get_inner_pool(
         &self,
-    ) -> Arc<tokio::sync::RwLock<HashMap<InstalledAppId, AppWebsocket>>> {
+    ) -> Arc<tokio::sync::RwLock<HashMap<InstalledAppId, AppWebsocketWithState>>> {
         self.app_clients.clone()
     }
 }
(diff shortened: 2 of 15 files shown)


src/error.rs
```diff
@@ -73,6 +80,9 @@ impl IntoResponse for HcHttpGatewayError {
                     )),
                 )
             }
+            HcHttpGatewayError::UpstreamUnavailable => {
```
ThetaSinner, line 83: Not handling `HolochainError` here, I think that's a job for the zome calls task

src/app_conn_pool.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// Call a function with an app client for the given installed app ID.
+    ///
+    /// This function takes care of reconnecting to the app client if the connection is lost. Your
+    /// function is free to operate on the app client without worrying about the connection state.
+    ///
+    /// Exactly one attempt is made to reconnect and re-run the provided callback. If there is a
+    /// second failure after reconnecting, the error is returned to the caller.
+    pub async fn call<T>(
```
ThetaSinner, line 46: I've some of the other functions public but this should be the only one we need to call when making zome calls

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        // Issue an app authentication token to allow us to connect a new client.
+        let issued = admin_ws
+            .issue_app_auth_token(IssueAppAuthenticationTokenPayload::for_installed_app_id(
+                installed_app_id.clone(),
+            ))
+            .await?;
+
+        // Presence of host must have been checked to get an admin connection
+        let host = self
```
ThetaSinner, line 175: This is thoroughly grim. This is why I was keen to store a `SocketAddr` on the configuration. I'd like to make that change as a follow up.

If you run this with `localhost` it will fail to make connections. It has to be an IP address at this point.

.github/workflows/test.yaml
```diff
@@ -28,7 +28,9 @@ jobs:
         run: |
           nix develop -c ./fixture/package.sh
       - name: Run tests
-        run: cargo test --no-fail-fast
+        env:
+          RUST_LOG: warn,holochain_http_gateway=info
+        run: cargo test --no-fail-fast -- --test-threads=1
```
ThetaSinner, line 33: Sweettests are so heavy, running them in parallel doesn't help. We could be a bit more selective if we used nexttest here and defined the websocket tests as needing to run in sequence but... bigger job than I want to take on this in PR.

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+    ) -> HcHttpGatewayResult<T> {
+        for _ in 0..2 {
+            let app_ws = self
+                .get_or_connect_app_client(installed_app_id.clone(), admin_websocket.clone())
+                .await?;
+            match execute(app_ws).await {
+                Ok(response) => {
+                    return Ok(response);
+                }
+                Err(HcHttpGatewayError::HolochainError(ConductorApiError::WebsocketError(e))) => {
```
mattyg, line 70: Cheaper to store the app auth token expiration and check it ourselves before making the websocket call? Error due to token expiration will happen frequently.

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+                    HcHttpGatewayError::ConfigurationError(format!("Invalid socket address: {}", e))
+                },
+            )?)
+            .try_set_header("Origin", HTTP_GW_ORIGIN)?;
+
+        // Create a websocket client configuration and lower the default timeout. We are connecting
+        // locally to a running Holochain. If requests take longer than 10s then we want to free up
+        // the HTTP gateway to handle other requests.
+        let mut config = WebsocketConfig::CLIENT_DEFAULT;
+        config.default_request_timeout = std::time::Duration::from_secs(10);
```
mattyg, line 196: - I assume we've already run `init` on the installed app, before it is called via this gateway (so this will never trigger wasm compliation). If not its too short of a timeout.
- I could imagine heavy zome calls taking longer than 10s, but maybe that's just a hard constraint we're setting for app devs
- Should this timeout be configured in the gateway env vars?

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+            .values()
+            .flat_map(|cell_infos| {
+                cell_infos.iter().flat_map(|cell_info| {
+                    match cell_info {
+                        CellInfo::Provisioned(provisioned) => Some(provisioned.cell_id.clone()),
+                        // TODO Provisioning of these wouldn't be dynamic, you'd have to
+                        //      restart the gateway or Holochain to get new credentials for
+                        //      new clones...
+                        //      See https://github.com/holochain/holochain/issues/4595
+                        // CellInfo::Cloned(clone_cell) => Some(clone_cell.cell_id.clone()),
```
mattyg, line 240: The way this is written, we still wouldn't get credentials for clones, even after restarting right?

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        Ok(app_ws)
+    }
+
+    async fn get_app_port(
+        &self,
+        installed_app_id: &InstalledAppId,
+        admin_ws: AdminWebsocket,
+    ) -> HcHttpGatewayResult<u16> {
+        if let Some(app_port) = self.cached_app_port.read().expect("Invalid lock").as_ref() {
```
mattyg, line 283: Should this be in its own scope to release the read lock?

mattyg (changes requested):
looks good -- couple comments inline

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+                    app_ws: app_ws.clone(),
+                    opened_at: Timestamp::now(),
+                });
+
+                app_ws
+            }
+        };
+
+        if app_client_lock.len() > self.configuration.max_app_connections as usize {
+            // Find and remove the oldest connection
```
matthme, line 127: I guess oldest connection does not necessarily mean "longest time since last used". It could be that the first that's being connected to is also the one that's being used most frequently...but that's probably just nit-pick and not relevant in practice.

src/app_conn_pool.rs
matthme: 10 seconds is indeed quite short. Especially if this gateway is supposed to support remote calls as well, these can easily take several 10s of seconds in some cases.

src/app_conn_pool.rs
matthme: I thought the same yes. The way it is right now, cloned cells are not supported categorically.
I'm also noticing that granting of functions is only by zome name and function name and not role name. So there would be no way currently to restrict calling a function for specific clones or roles only.

tests/websocket.rs
```diff
@@ -20,6 +38,409 @@ async fn connect_app_websocket() {
...
+            .unwrap()
+            .installed_app_id
+    );
+
+    // Demonstrate that the pool was prevented from writing by the read lock held above.
+    assert!(inner_pool.try_write().is_err());
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn does_not_reconnect_on_non_websocket_error() {
```
matthme, line 138: I don't understand what this test tests. Would it be possible to add a bit more description beyond what the function name says?

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+    ///
+    /// This function takes care of reconnecting to the app client if the connection is lost. Your
+    /// function is free to operate on the app client without worrying about the connection state.
+    ///
+    /// Exactly one attempt is made to reconnect and re-run the provided callback. If there is a
+    /// second failure after reconnecting, the error is returned to the caller.
+    pub async fn call<T>(
+        &self,
+        installed_app_id: InstalledAppId,
+        admin_websocket: AdminWebsocket,
```
jost-s, line 59: Once the admin websocket PR is merged, perhaps it would make sense to add admin ws management to this pool instead of injecting it?

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+/// A connection pool for app connections.
+///
+/// This is a pool in the sense that it manages multiple connections to Holochain app interfaces,
+/// but it will manage exactly one connection per installed app.
+#[derive(Debug, Clone)]
+pub struct AppConnPool {
+    configuration: Configuration,
+    cached_app_port: Arc<RwLock<Option<u16>>>,
```
jost-s, line 35: Is this like reserving an app port?

jost-s (commented):
Looks great! Some comments

tests/websocket.rs
matthme: Ah okay, because `get_or_connect_app_client` wouldn't be able to get the lock :+1: . I don't understand yet though why the zome call produces an error at all, it looks valid to me.

src/app_conn_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+
+                // If we failed to make a connection, clear the cached app port so that the next
+                // attempt will re-check the app interfaces.
+                *self.cached_app_port.write().expect("Invalid lock") = None;
+
+                // Try again, with one fewer retry permitted
+                return Box::pin(self.attempt_connect_app_ws(
+                    installed_app_id,
+                    admin_ws,
+                    retries - 1,
```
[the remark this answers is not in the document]
ThetaSinner: Oh dear, yes I moved this code around too many times. I've removed this level of retry and handled the retry logic up in the `call` function for both disconnected websockets and failure to connect.

I also added a new test that attempting to reconnect returns an error and doesn't just hang -> `reconnect_gives_up`

https://github.com/holochain/hc-http-gw/pull/17/commits/394963ae93978fd098e03ddf63faefbe7da4d050

src/app_conn_pool.rs
ThetaSinner: We're not caching the token and the token lifetime is only used by Holochain to check that it is still valid when the connection is made, there's no need to reconnect later with a new token if the websocket stays open. Because we're using single-use tokens, I don't think it should really happen that we fail with a token error. It's more likely that Holochain closed the connection on a conductor restart or something like that. We might fail to connect if we used a cached port, and hopefully the updated code makes that clearer. Otherwise if we get connection error without anything cached, there's no reason to retry connecting.

https://github.com/holochain/hc-http-gw/pull/17/commits/394963ae93978fd098e03ddf63faefbe7da4d050

src/app_conn_pool.rs
ThetaSinner: Yes, running `init` is a pre-requisite and the tests have to do that too.
Possible 10s is too low and I agree we should make this a configuration option. Making remote calls is not expected behavior though, and nor is going to the network ideally. This is supposed to serving DHT data and not also fetching it.

https://github.com/holochain/hc-http-gw/pull/17/commits/6d038a3310a574646592df77399e4243bb1c37e4

src/app_conn_pool.rs
ThetaSinner: You're absolutely right, and that would be a cleaner heuristic for dropping connections. The trouble with it would be that we'd need to write the last used time somewhere which involves synchronizing writes. I've tried to use RwLocks and prefer doing reads where possible. Those can happen in parallel unless there is a write needing to happen. That should perform better than just a mutex that is always a bottleneck.

Doing it this way will create churn if we hit the connection limit. I think its fairly unlikely to happen and we're in a bad state if we reach the limit anyway really because we're trying to support more apps than we planned for. We might end up reconnecting some apps that are being actively used but that should just rotate the list and the ones that aren't being used won't get re-opened. So the effect is the same with a bit of extra noise. If we are allowing 50 connections and somebody is trying to host 55 apps, there's not much we can do to be efficient either way.

tests/websocket.rs
jost-s: Yes, that warrants a comment.

jost-s: Or perhaps the closure can just return an error instead of making a call?

tests/websocket.rs
```diff
@@ -20,6 +38,409 @@ async fn connect_app_websocket() {
...
+    // Should have provisioned exactly 1 app interface for the HTTP gateway
+    //
+    // Note that this check would also pass if the conductor was exposing an app interface with
+    // allowed origins *.
+    let app_interfaces = sweet_conductor.list_app_interfaces().await.unwrap();
+    let matched_app_interfaces = app_interfaces
+        .iter()
+        .filter(|interface| interface.allowed_origins.is_allowed(HTTP_GW_ORIGIN))
+        .collect::<Vec<_>>();
+    assert_eq!(matched_app_interfaces.len(), 1);
```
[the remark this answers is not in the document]
ThetaSinner: One interface, two connections. So we should create one compatible app interface and connect to it twice. On the other end, it's just a TcpListener and each connection can be handled separately anyway.

src/app_conn_pool.rs
ThetaSinner: Agreed, this needs updating to integrate with the admin websocket PR

src/app_conn_pool.rs
jost-s: It's all local, so the listing call should be fast. Fast enough that I'd prefer to not cache it, but I'm not opposed if you consider this important.

jost-s (approved):
Super

matthme (approved):

src/app_conn_pool.rs
ThetaSinner: Oh I see what you're saying, yes that's a good point. So you could either permit access to all clones if the DNA hash is known or you'd have to add more configuration to the gateway to restrict which clones are permitted.

src/app_conn_pool.rs
ThetaSinner: It might be a little heavy on optimisation. I'm a little concerned about this running alongside Envoy which has a bad habit of allocating a silly number of app interfaces - but that is going away...

I'd like to keep this for now, it's trivial to remove.

src/app_conn_pool.rs
```diff
@@ -0,0 +1,334 @@
...
+use holochain_client::{
+    AdminWebsocket, AppWebsocket, AuthorizeSigningCredentialsPayload, CellInfo, ClientAgentSigner,
+    ConductorApiError, ConnectRequest, GrantedFunctions, IssueAppAuthenticationTokenPayload,
+    Timestamp, WebsocketConfig,
+};
+use holochain_types::app::InstalledAppId;
+use holochain_types::websocket::AllowedOrigins;
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::sync::{Arc, RwLock};
```
c12i, line 13: Could we use `tokio::sync::RwLock` for consistency?

c12i (changes requested):
One comment, otherwise looks good

src/app_conn_pool.rs
ThetaSinner: I don't think consistency is the goal here. I'd prefer to use the stdlib but the `app_clients` lock has to be held across `.await` points. Is there a benefit to using a tokio RwLock instead of the stdlib one?

c12i: I asked because I noticed that the `cached_app_port` is under the std lib `RwLock` while the rest are on tokio's, checking if this was intentional.

c12i (approved):

PR #18, opened 2025-03-06
chore: Add test fixture

ThetaSinner:


fixture/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
+members = [
```
ThetaSinner, line 2: One integrity zome and two coordinators so that we can set up tests which have the same DNA hash but are actually different apps

tests/websocket.rs
```diff
@@ -0,0 +1,25 @@
+use crate::sweet::{install_fixture1, install_fixture2};
+use holochain::sweettest::SweetConductor;
+use holochain_client::AdminWebsocket;
+use std::net::Ipv4Addr;
+
+mod sweet;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn connect_app_websocket() {
```
ThetaSinner, line 9: Not a very interesting test, just to prove the fixtures are ready to use

fixture/package.sh
```diff
@@ -0,0 +1,23 @@
...
+hc dna pack .
+popd || exit 1
+
+pushd package/happ1 || exit 1
+hc app pack .
+popd || exit 1
+
+pushd package/happ2 || exit 1
+hc app pack .
+popd || exit 1
```
cdunster, line 23: Nitpicks:
1. The cargo build lines could be combined by adding multiple `-p` flags.
2. This script, as is, needs to be run from inside the fixtures directory, but you can use `dirname $0` to make it runnable anywhere.

```suggestion
SCRIPT_DIR=$(dirname "$0")

cargo build --manifest-path "$SCRIPT_DIR/Cargo.toml" --release --target wasm32-unknown-unknown -p integrity -p coordinator1 -p coordinator2

pushd "$SCRIPT_DIR/package/dna1" || exit 1
hc dna pack .
popd || exit 1

pushd "$SCRIPT_DIR/package/dna2" || exit 1
hc dna pack .
popd || exit 1

pushd "$SCRIPT_DIR/package/happ1" || exit 1
hc app pack .
popd || exit 1

pushd "$SCRIPT_DIR/package/happ2" || exit 1
hc app pack .
popd || exit 1

```

cdunster (commented):
Just some suggestions for the package script.

fixture/package.sh
cdunster: Ah, well done, GitHub. Great diff :roll_eyes:

ThetaSinner: Nice, yes that works well thank you :) Also tweaked CI to call the script from the root of the repository

flake.nix
```diff
@@ -2,40 +2,42 @@
...
-          inherit system;
-          overlays = [ inputs.rust-overlay.overlays.default ];
+      formatter = pkgs.nixpkgs-fmt;
+
+      devShells.default =
+        let
+          rustFromFile = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+        in
+        pkgs.mkShell {
+          packages = (with inputs'.holonix.packages; [
```
c12i, line 27: I think we need `hc` here, right? The fixture script calls `hc app` and `hc dna`

c12i (approved):

flake.nix
ThetaSinner: That's included with `holochain`, we don't export the binaries from the Holochain repository one by one at the moment in Holonix. It's all one Nix package with several binaries

PR #19, opened 2025-03-07
feat: implement app selection logic

### LINKED ISSUE #9: [httpgw] Implement app selection logic
The gateway should manage a single admin connection. Opening app connections is not required for this issue.

The app selection logic should be implemented based on the logic described in the spec. By taking the incoming request URL, it should be possible to identify a unique app that a request can be made to later.

AC:
- First call admin_ws.list_apps and filter list to get target app; keep response cached for subsequent calls
  - 404 Response for an app that is not in the list of installed apps
- 403 Response if the request is not permitted: check if within HC_GW_ALLOWED_APP_IDS & HC_GW_ALLOWED_FNS_<app-id>
- App selection logic implemented
- Selected app id is printed for now
- Tests: app selection logic

cdunster:
Requires an actual `AdminWebsocket` that will be implemented with #11.

closes #9

### DIFF
--- a/src/app_selection.rs
+++ b/src/app_selection.rs
@@ -1,0 +1,82 @@
+use holochain_client::AppInfo;
+use holochain_types::dna::DnaHashB64;
+use thiserror::Error;
+
+#[derive(Debug, PartialEq, Error)]
+pub enum AppSelectionError {
+    #[error("App is not installed on the conductor")]
+    NotInstalled,
+}
+
+fn check_app_valid(
+    dna_hash: DnaHashB64,
+    installed_apps: &[AppInfo],
+) -> Result<(), AppSelectionError> {
+    installed_apps
+        .iter()
+        .any(|a| {
+            a.manifest.app_roles().iter().any(|r| {
+                r.dna
+                    .installed_hash
+                    .as_ref()
+                    .is_some_and(|hash| hash == &dna_hash)
+            })
+        })
+        .then_some(())
+        .ok_or(AppSelectionError::NotInstalled)
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use super::*;
+    use assert2::assert;
+    use holochain::core::DnaHash;
+    use holochain_client::AgentPubKey;
+    use holochain_types::app::{AppManifest, AppRoleDnaManifest, AppRoleManifest, AppStatus};
+
+    fn new_fake_app_info(app_id: impl ToString, dna_hash: DnaHashB64) -> AppInfo {
+        AppInfo {
+            installed_app_id: app_id.to_string(),
+            cell_info: HashMap::new(),
+            status: AppStatus::Running.into(),
+            agent_pub_key: AgentPubKey::from_raw_32([1; 32].to_vec()),
+            manifest: AppManifest::V1(holochain_types::app::AppManifestV1 {
+                name: Default::default(),
+                description: Default::default(),
+                roles: vec![AppRoleManifest {
+                    name: Default::default(),
+                    provisioning: Default::default(),
+                    dna: AppRoleDnaManifest {
+                        location: Default::default(),
+                        modifiers: Default::default(),
+                        installed_hash: Some(dna_hash),
+                        clone_limit: Default::default(),
+                    },
+                }],
+                allow_deferred_memproofs: Default::default(),
+            }),
+        }
+    }
+
+    #[test]
+    fn returns_error_if_app_not_installed() {
+        let dna_hash = DnaHash::from_raw_32([1; 32].to_vec()).into();
+        let installed_apps = [];
+
+        let result = check_app_valid(dna_hash, &installed_apps);
+
+        assert!(result == Err(AppSelectionError::NotInstalled));
+    }
+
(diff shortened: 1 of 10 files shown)


src/app_selection.rs
```diff
@@ -0,0 +1,352 @@
...
+    dna_hash: &DnaHash,
+    coordinator_identifier: &str,
+    installed_apps: &'a [AppInfo],
+) -> Result<&'a AppInfo, AppSelectionError> {
+    let mut found_apps = installed_apps.iter().filter(|a| {
+        // TODO: Use real `coordinator_identifier` when field available.
+        a.installed_app_id == coordinator_identifier
+            && a.manifest.app_roles().iter().any(|r| {
+                r.dna
+                    .installed_hash
```
ThetaSinner, line 31: Is installed_hash expected to be present? I'd expect to have to look at the `cell_info` to get the DNA hashes in general

src/app_selection.rs
```diff
@@ -0,0 +1,352 @@
...
+                    .as_ref()
+                    .is_some_and(|hash| &Into::<DnaHash>::into(hash.clone()) == dna_hash)
+            })
+    });
+
+    let app_info = found_apps.next().ok_or(AppSelectionError::NotInstalled)?;
+
+    if found_apps.next().is_some() {
+        return Err(AppSelectionError::MultipleMatching);
```
ThetaSinner, line 40: Log a warning for the administrator to fix their mistake before a user has to report it?

src/app_selection.rs
```diff
@@ -0,0 +1,268 @@
...
+    } else {
+        *installed_apps = admin_websocket.list_apps();
+        find_installed_app(&dna_hash, installed_apps).ok_or(AppSelectionError::NotInstalled)?
+    };
+
+    allowed_apps
+        .contains(&app_info.installed_app_id)
+        .then_some(())
+        .ok_or(AppSelectionError::NotAllowed)?;
+
```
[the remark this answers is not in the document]
ThetaSinner: Agree, let the zome call code check that for the chosen app

src/app_selection.rs
```diff
@@ -0,0 +1,352 @@
...
+    {
+        app_info
+    } else {
+        *installed_apps = admin_websocket.list_apps().await.unwrap();
+        find_installed_app(&dna_hash, &coordinator_identifier, installed_apps)?
+    };
+
+    allowed_apps
+        .contains(&app_info.installed_app_id)
+        .then_some(())
```
ThetaSinner, line 72: This is fancy :) 

Alternatively and slightly faster to decode?

```rust
if !allowed_apps.contains(&app_info.installed_app_id) {
    return Err(AppSelectionError::NotAllowed);
}
```

src/app_selection.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    use super::*;
+    use assert2::assert;
+    use holochain_client::AgentPubKey;
+    use holochain_types::app::{AppManifest, AppRoleDnaManifest, AppRoleManifest, AppStatus};
+
+    fn new_fake_app_info(app_id: impl ToString, dna_hash: DnaHash) -> AppInfo {
+        AppInfo {
+            installed_app_id: app_id.to_string(),
+            cell_info: HashMap::new(),
```
ThetaSinner, line 92: This will always be filled in for a conductor response

src/router.rs
```diff
@@ -1,7 +1,7 @@
 use crate::holochain::AppCall;
 use crate::{
     config::Configuration,
-    routes::{health_check, zome_call},
+    routes::{app_selection, health_check, zome_call},
```
ThetaSinner, line 4: under routes? Maybe this belongs as its own top level module?

src/router.rs
```diff
@@ -17,9 +17,11 @@ pub fn hc_http_gateway_router(
         configuration,
         admin_call,
         app_call,
+        installed_apps: Default::default(),
     };
     Router::new()
         .route("/health", get(health_check))
+        .route("/{dna_hash}/{coordinator_identifier}", get(app_selection))
```
ThetaSinner, line 24: Oh I see, we're adding something new. Is this just for now for testing?

src/service.rs
```diff
@@ -15,7 +16,7 @@ pub struct HcHttpGatewayService {
 }
 
 /// Shared application state
-#[derive(Debug, Clone)]
+#[derive(Clone)]
```
ThetaSinner, line 19: Do we need to implement debug for `AppInfo`?

ThetaSinner: If so, we might want to wrap the new field in a newtype and implement debug for that for now

ThetaSinner (changes requested):
A few comments. I think the main thing is to not rely on `installed_hash`. That's an optional field and I think it's incredibly confusing. It's designed to allow somebody to package their app with properties that they expect to be included in the DNA hash. If somebody changes those properties or adds a network seed without updating the `installed_hash` then the app will fail to install on Holochain. It's a check that somebody really meant to end up on a different network. Its use is optional and I suspect not common.

ThetaSinner:
Continued on #28

Issue #12, opened 2025-02-13
[httpgw] Implement zome calls


ThetaSinner:
With the logic in place to receive HTTP requests, select an app, and make an app websocket connection - it should be possible to make zome calls.

AC:
- Zome calls are made to Holochain when an HTTP GET requests is made that can be translated to a valid zome call.
- Ensure that the zome fn is part of HC_GW_ALLOWED_FNS_<app_id>
- Basic test written (e.g. using reqwest and sweettest)
- Convert JSON payload to msgpack and the result from the zome call to JSON which is sent back to the user
- Error handling

PR #22, opened 2025-03-09
Refactor zome call path validations

jost-s:
* Adds transcode functions to convert from base64 JSON -> ExternIO and ExternIO -> JSON
* Adds router module for simpler testability of routes
* Adds/refactors path validations of zome call route
* Add fallback to router for unauthorized request methods

This is part of #12, but not complete. The next PR will contain the actual zome call and error handling.


src/error.rs
```diff
@@ -45,51 +49,61 @@ impl IntoResponse for HcHttpGatewayError {
...
-/// Construct an axum http error from a status code and optional message
-pub fn error_response(status_code: u16, message: Option<&str>) -> axum::response::Response {
-    let error_response = HcGwErrorResponse {
-        error: message.unwrap_or("Something Went Wrong").to_string(),
-    };
-
-    (
-        StatusCode::from_u16(status_code).unwrap_or(StatusCode::INTERNAL_SERVER_ERROR),
-        Json(error_response),
-    )
```
jost-s, line 93: A slight duplication with the catchall case above and the unwrap_ors here.

src/http.rs
jost-s, line 1: Returning a plain string as error instead of this type with an error field.

src/router.rs
jost-s, line 1: Following the [example how to test the router](https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs#L37), this is a builder for the router which can then be tested.

src/routes/zome_call.rs
```diff
@@ -55,53 +77,127 @@ pub async fn zome_call(
...
+        let _zome_call_payload = base64_json_to_hsb(&payload)?;
     }
 
     Ok(())
 }
 
-/// Calculate the approximate decoded size without actually decoding
-/// Base64 encoding: every 4 chars in base64 represent 3 bytes of original data
-/// Need to account for padding characters too ('='), which don't represent data
-fn calculate_base64_decoded_size(encoded_payload: &str) -> u32 {
```
jost-s, line 94: Actually I think now that we need to check the byte length of the base64 encoded payload, not the length of the decoded payload. The spec says this limit is supposed to restrict the overall request size.

src/transcode.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    #[test]
+    fn invalid_json_to_hsb_fails() {
+        let base64_encoded_payload = BASE64_URL_SAFE.encode("invalid");
+
+        let result = base64_json_to_hsb(&base64_encoded_payload);
+        assert!(matches!(result, Err(HcHttpGatewayError::InvalidJSON(_))));
+    }
+
```
jost-s, line 82: I couldn't come up with a (reasonable) test to produce a failing `ExternIO` encoding. [`SerializedBytesError`](https://github.com/holochain/holochain-serialization/blob/main/crates/holochain_serialized_bytes/src/lib.rs#L64-L71) also considers it very unlikely.

src/router.rs
```diff
@@ -0,0 +1,84 @@
+use crate::{
+    config::Configuration,
+    routes::{health_check, zome_call},
+    service::AppState,
+};
+use axum::{routing::get, Router};
+
+pub fn hc_http_gateway_router(configuration: Configuration) -> Router {
```
jost-s, line 8: Following the [example how to test the router](https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs#L37), this is a builder for the router which can then be tested.

src/http.rs
jost-s, line 1: Moved to error.rs

tests/zome_call.rs
```diff
@@ -24,13 +24,12 @@ async fn zome_call_with_valid_params() {
     let response = app
         .call_zome(
             &dna_hash,
-            "coord98765",
+            "forum",
```
jost-s, line 27: These tests were calling unauthorized functions, but there was no validation yet, so they passed before.

src/config.rs
```diff
@@ -344,6 +361,49 @@ mod tests {
...
+            assert_eq!(
+                config.is_function_allowed("nopp", "zome_name", "fn_name"),
+                false
+            );
+        }
+
+        #[test]
+        fn is_function_allowed_returns_true_when_all_functions_allowed_for_app() {
+            let config = create_test_config();
+            println!("allwoed fns {:?}", config.allowed_fns);
```
matthme, line 376:
```suggestion
            println!("allowed fns {:?}", config.allowed_fns);
```

src/config.rs
```diff
@@ -344,6 +361,49 @@ mod tests {
...
+        fn is_function_allowed_returns_false_when_app_is_not_found() {
+            let config = create_test_config();
+            assert_eq!(
+                config.is_function_allowed("nopp", "zome_name", "fn_name"),
+                false
+            );
+        }
+
+        #[test]
+        fn is_function_allowed_returns_true_when_all_functions_allowed_for_app() {
```
matthme, line 374: I'm noticing when reading this test, that the `is_function_allowed` method does not check whether a function exists at all for a given app. That's probably okay in the contexts in which the method is being used but I thought I'd mention it anyway.

src/transcode.rs
```diff
@@ -0,0 +1,100 @@
...
+        // Create a payload, serialize to JSON and base64 encode.
+        let payload = ZomeCallPayload { field: false };
+        let json_payload = serde_json::to_string(&payload).unwrap();
+        let base64_encoded_payload = BASE64_URL_SAFE.encode(json_payload);
+
+        let hsb_encoded_payload = base64_json_to_hsb(&base64_encoded_payload).unwrap();
+
+        // Deserializing the serialized bytes to the original struct should succeed.
+        let decoded_payload = hsb_encoded_payload.decode::<ZomeCallPayload>().unwrap();
+        assert_eq!(decoded_payload, payload);
```
ThetaSinner, line 55: Nice!

ThetaSinner (commented):
This is looking good. The added tests in particular.

My main concerns at the moment are:
- How we're handling errors
- That the new unit tests will become integration tests with the addition of connections to Holochain. I wonder if you've already got a solution to that or whether we might end up having to rethink how the code is split up for testing.

src/router.rs
jost-s: I think we should still be able to test path extraction and validation separately. So I suggest to make the app connection pool mockable? Then we can test with state but don't need a Holochain connection.

src/config.rs
```diff
@@ -344,6 +361,49 @@ mod tests {
             assert!(config.get_allowed_functions("app3").is_none());
         }
 
+        #[test]
+        fn is_function_allowed_returns_false_when_app_is_not_found() {
+            let config = create_test_config();
+            assert_eq!(
+                config.is_function_allowed("nopp", "zome_name", "fn_name"),
+                false
+            );
```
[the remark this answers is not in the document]
jost-s: This is fine for me too and I think clippy even demands since some version. Usually I wasn't happy with how failed assertions were logged when using `assert`, but in this case it's fine.

src/error.rs
```diff
@@ -3,93 +3,162 @@
...
+        fn_name: String,
+    },
 }
 
 /// Type aliased Result
 pub type HcHttpGatewayResult<T> = Result<T, HcHttpGatewayError>;
 
+/// Error format returned to the caller.
+#[derive(Debug, Serialize)]
+pub struct JsonErrorResponse {
```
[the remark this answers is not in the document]
jost-s: I had named it `ErrorResponse` at first, but that's also a type in axum. But that's not a problem. Is that fine with you, `ErrorResponse`?

src/routes/zome_call.rs
```diff
@@ -55,53 +83,171 @@ pub async fn zome_call(
...
+        .is_function_allowed(&coordinator_identifier, &zome_name, &fn_name)
+    {
+        return Err(HcHttpGatewayError::UnauthorizedFunction {
+            app_id: coordinator_identifier,
+            zome_name,
+            fn_name,
+        });
+    }
+    // Check payload byte length does not exceed configured maximum.
+    if let Some(payload) = query.payload {
```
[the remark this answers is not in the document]
jost-s: It's a hashmap lookup, but fine =)

src/transcode.rs
```diff
@@ -0,0 +1,100 @@
...
+//! to a JSON string.
+
+use crate::HcHttpGatewayResult;
+use base64::{prelude::BASE64_URL_SAFE, Engine};
+use holochain_types::prelude::ExternIO;
+
+/// Function to transcode an incoming base64 encoded payload to Holochain serialized bytes
+/// (type `ExternIO`).
+pub fn base64_json_to_hsb(base64_encoded_payload: &str) -> HcHttpGatewayResult<ExternIO> {
+    let base64_decoded_payload = BASE64_URL_SAFE.decode(base64_encoded_payload)?;
```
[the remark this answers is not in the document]
jost-s: This only affects the caller who sends a request to the gateway. They need to encode their zome call payload using this encoding. In the gateway spec it mentions the URL safe method, so we should be good.

src/config.rs
jost-s: done https://github.com/holochain/hc-http-gw/pull/22/commits/2c616464a7e4db043fa0baffea78e555084a18d7

src/config.rs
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/hc-http-gw/pull/22/commits/2c616464a7e4db043fa0baffea78e555084a18d7

src/routes/zome_call.rs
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/hc-http-gw/pull/22/commits/cbcab1f47f6339ac61e0c23951bdbfa677dd79d7

src/error.rs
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/hc-http-gw/pull/22/commits/b1c2f2de7cb851ee7e0958405d43a886db990b5d

src/config.rs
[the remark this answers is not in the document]
ThetaSinner: Looks good. There are a few more assertions like this one in the same file if that's alright?

src/router.rs
ThetaSinner: Sounds good, we'll figure that out as we integrate these PRs

ThetaSinner (approved):
Looks good! Could collapse some more `assert_eq!` with booleans but all changes looking good

src/config.rs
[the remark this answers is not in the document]
jost-s: Ah sure, I'll do them too.

jost-s:
Here we go https://github.com/holochain/hc-http-gw/pull/22/commits/2cc569976c8a4a9eb98419179346cc84f08b5854

ThetaSinner (approved):

PR #24, opened 2025-03-10
fix: Improve log readability

### LINKED ISSUE #20: fix: Improve logs readability
Current logs are in JSON by default and not easily readably unless piped to a tool like [bunyan]

AC:

- Have the logs easily readable out the box

c12i:
Closes #20


ThetaSinner (approved):
Looks good locally. I think we might want to put this in the binary and have the tests set up their own default level. It'd be nice to have `holochain=warn,info` by default so that running `cargo test` doesn't produce as much output by default.

PR #25, opened 2025-03-11
refactor(error)!: simplify error responses

jost-s:
Combine all errors related to request path validation and transcoding under one error variant and remove error tracing for them.

### DIFF
--- a/src/app_conn_pool.rs
+++ b/src/app_conn_pool.rs
@@ -197,13 +197,13 @@ impl AppConnPool {
         tracing::info!("Connecting to app websocket at {}:{}", host, app_port);
 
         // Build a connection request
-        let request =
-            ConnectRequest::from(format!("{host}:{app_port}").parse::<SocketAddr>().map_err(
-                |e| {
-                    HcHttpGatewayError::ConfigurationError(format!("Invalid socket address: {}", e))
-                },
-            )?)
-            .try_set_header("Origin", HTTP_GW_ORIGIN)?;
+        let request = ConnectRequest::from(
+            format!("{host}:{app_port}")
+                .parse::<SocketAddr>()
+                .expect("Host and app port must be valid"),
+        )
+        .try_set_header("Origin", HTTP_GW_ORIGIN)
+        .expect("Origin headers have gone out of fashion");
 
         // Create a websocket client configuration and lower the default timeout. We are connecting
         // locally to a running Holochain. If requests take longer than the configured timeout then
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,10 +8,7 @@ use std::{
     ops::Deref,
     str::FromStr,
 };
-
 use url2::Url2;
-
-use crate::{HcHttpGatewayError, HcHttpGatewayResult};
 
 /// Default payload size limit (10 kilobytes)
 pub const DEFAULT_PAYLOAD_LIMIT_BYTES: u32 = 10 * 1024;
@@ -21,6 +18,23 @@ pub const DEFAULT_MAX_APP_CONNECTIONS: u32 = 500;
 
 /// Default timeout for zome calls
 pub const DEFAULT_ZOME_CALL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
+/// Errors when parsing config arguments.
+#[derive(Debug, thiserror::Error)]
+pub enum ConfigParseError {
+    /// Error when parsing URL.
+    #[error("URL parse error: {0}")]
+    UrlParseError(#[from] url2::Url2Error),
+    /// Error when parsing an integer.
+    #[error("Integer parse error: {0}")]
+    IntParseError(#[from] std::num::ParseIntError),
+    /// Other parsing error.
+    #[error("Parse error: {0}")]
+    Other(String),
+}
+
+/// Result of parsing config arguments.
+pub type ConfigParseResult<T> = std::result::Result<T, ConfigParseError>;
 
 /// Main configuration structure for the HTTP Gateway.
 #[derive(Debug, Clone)]
@@ -55,27 +69,20 @@ impl Configuration {
         allowed_fns: HashMap<AppId, AllowedFns>,
         max_app_connections: &str,
         zome_call_timeout: &str,
-    ) -> HcHttpGatewayResult<Self> {
-        let admin_ws_url = Url2::try_parse(admin_ws_url).map_err(|e| {
-            HcHttpGatewayError::ConfigurationError(format!("Url parse error: {}", e))
-        })?;
+    ) -> ConfigParseResult<Self> {
+        let admin_ws_url = Url2::try_parse(admin_ws_url)?;
 
         let payload_limit_bytes = if payload_limit_bytes.is_empty() {
             DEFAULT_PAYLOAD_LIMIT_BYTES
         } else {
-            payload_limit_bytes.parse::<u32>().map_err(|e| {
-                HcHttpGatewayError::ConfigurationError(format!(
-                    "Failed to parse the payload limit bytes value: {}",
-                    e
-                ))
-            })?
+            payload_limit_bytes.parse::<u32>()?
         };
 
         let allowed_app_ids = AllowedAppIds::from_str(allowed_app_ids)?;
 
         for app_id in allowed_app_ids.iter() {
             if !allowed_fns.contains_key(app_id) {
-                return Err(HcHttpGatewayError::ConfigurationError(format!(
+                return Err(ConfigParseError::Other(format!(
                     "{} is not present in allowed_fns",
                     app_id
                 )));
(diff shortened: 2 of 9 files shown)


src/config.rs
```diff
@@ -55,53 +53,33 @@ impl Configuration {
         allowed_fns: HashMap<AppId, AllowedFns>,
         max_app_connections: &str,
         zome_call_timeout: &str,
-    ) -> HcHttpGatewayResult<Self> {
-        let admin_ws_url = Url2::try_parse(admin_ws_url).map_err(|e| {
-            HcHttpGatewayError::ConfigurationError(format!("Url parse error: {}", e))
-        })?;
+    ) -> anyhow::Result<Self> {
```
jost-s, line 56: I think if the configuration cannot be created, the error can be returned as is as a string.

src/service.rs
```diff
@@ -27,7 +24,7 @@ impl HcHttpGatewayService {
         address: impl Into<IpAddr>,
         port: u16,
         configuration: Configuration,
-    ) -> HcHttpGatewayResult<Self> {
+    ) -> std::io::Result<Self> {
```
jost-s, line 27: If the service cannot be run because of IO errors, it can be returned as those.

src/app_conn_pool.rs
```diff
@@ -197,13 +197,13 @@ impl AppConnPool {
...
-                },
-            )?)
-            .try_set_header("Origin", HTTP_GW_ORIGIN)?;
+        let request = ConnectRequest::from(
+            format!("{host}:{app_port}")
+                .parse::<SocketAddr>()
+                .expect("Host and app port must be valid"),
+        )
+        .try_set_header("Origin", HTTP_GW_ORIGIN)
+        .expect("Origin headers have gone out of fashion");
```
jost-s, line 206: This looks pretty bold, but we're already expecting the host address earlier and the port is also assured. Regarding headers, "Origin" is always safe to set. Or do you prefer to re-introduce std::io::Errors in the error enum and return a 500 error to callers (in theory, because it'll never happen in practice)?

src/transcode.rs
```diff
@@ -77,7 +82,10 @@ mod tests {
         let base64_encoded_payload = BASE64_URL_SAFE.encode("invalid");
 
         let result = base64_json_to_hsb(&base64_encoded_payload);
-        assert!(matches!(result, Err(HcHttpGatewayError::InvalidJSON(_))));
+        assert!(matches!(
+            result,
+            Err(HcHttpGatewayError::RequestMalformed(err)) if err == "Invalid JSON value".to_string()
+        ));
```
cdunster, line 88: I personally like using the [assert2](https://crates.io/crates/assert2) crate for these kinds of checks and I've added it on my branch but here it could be written as:
```rust
use assert2::{let_assert, assert};

let_assert!(Err(HcHttpGatewayError::RequestMalformed(err)) = result);
assert!(err.to_string() == "Invalid JSON value"); // N.B. this is not the normal `assert`
```
where the first line checks the error type and assigns `err` which is then used on the second line to check the error message. It produces nicer outputs than just "Assert evaluated to `false`".

src/config.rs
```diff
@@ -3,16 +3,14 @@
 //! This module provides the configuration structure and related types for
 //! controlling the behavior of the HTTP Gateway.
 
+use anyhow::anyhow;
```
ThetaSinner, line 6: Please can we keep `anyhow` out of the library code. It's great when there are no consumers but trying to include anyhow errors into any other error handling mechanism than itself is a right pain.

src/app_conn_pool.rs
ThetaSinner: This is a step in the right direction. I think we don't need to be messing with strings at all at this point at all. We should resolve to an IP address as soon as we get the input configuration and then just be passing a Holochain IP around instead of strings. I've commented about this on a couple of PRs now but don't want to change too much at once :D

src/config.rs
ThetaSinner: Solid case for an "Other". Nothing you can do in response so an other with a context string does seem appropriate

src/routes/zome_call.rs
```diff
@@ -91,11 +91,11 @@ pub async fn zome_call(
     } = params;
     // Check payload byte length does not exceed configured maximum.
     if let Some(payload) = &query.payload {
-        if payload.as_bytes().len() > state.configuration.payload_limit_bytes as usize {
-            return Err(HcHttpGatewayError::PayloadSizeLimitError {
-                size: payload.as_bytes().len() as u32,
-                limit: state.configuration.payload_limit_bytes,
-            });
+        if payload.len() > state.configuration.payload_limit_bytes as usize {
```
ThetaSinner, line 94: Hmm, this is correct I think but it's not that clear why. Because we're accepting base64, we know there are no multi-byte UTF characters. Could you maybe leave a comment here that the string length is the same as the byte length because we know the format of the string?

tests/websocket.rs
```diff
@@ -502,7 +500,7 @@ async fn close_old_connections_on_limit() {
     assert_eq!(ws_for_apps, vec!["app_1", "app_3"]);
 }
 
-fn create_test_configuration(admin_port: u16) -> HcHttpGatewayResult<Configuration> {
+fn create_test_configuration(admin_port: u16) -> anyhow::Result<Configuration> {
```
ThetaSinner, line 503: Nice, yes in tests it's helpful

ThetaSinner (changes requested):
Really like how much the error enum has been slimmed down. Just want to tweak `anyhow` out of the library code.

src/config.rs
jost-s: I suppose you meant `std::io::Error::other`. The configuration `try_new` is only used when taking all the input from the console and parsing it https://github.com/holochain/hc-http-gw/pull/25/commits/e9c81a55b96d531775aab730e954f988136b34dc

src/config.rs
```diff
@@ -55,53 +53,33 @@ impl Configuration {
...
 
         let allowed_app_ids = AllowedAppIds::from_str(allowed_app_ids)?;
 
         for app_id in allowed_app_ids.iter() {
             if !allowed_fns.contains_key(app_id) {
-                return Err(HcHttpGatewayError::ConfigurationError(format!(
-                    "{} is not present in allowed_fns",
-                    app_id
-                )));
+                return Err(anyhow!("{} is not present in allowed_fns", app_id));
```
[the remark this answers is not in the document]
jost-s: `anyhow` removed from this module

src/config.rs
```diff
@@ -128,7 +106,7 @@ impl Deref for AllowedAppIds {
 }
 
 impl FromStr for AllowedAppIds {
-    type Err = HcHttpGatewayError;
+    type Err = anyhow::Error;
```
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/hc-http-gw/pull/25/commits/e9c81a55b96d531775aab730e954f988136b34dc

src/transcode.rs
jost-s: Cool, I'm not thrilled about the output `assert` produced, but `let_assert` is quite handy indeed. This produces something equally informative, but with the default formatting https://github.com/holochain/hc-http-gw/pull/25/commits/fa3b22d7ee4d51c3988780ff14d5e63f7b77939e

src/config.rs
jost-s: https://github.com/holochain/hc-http-gw/pull/25/commits/e9c81a55b96d531775aab730e954f988136b34dc

src/routes/zome_call.rs
jost-s: Added comments and fixed the char calc https://github.com/holochain/hc-http-gw/pull/25/commits/d06199f68a6e3ceaa7242d246b1ca36c0ae1b439

src/config.rs
```diff
@@ -128,11 +111,11 @@ impl Deref for AllowedAppIds {
 }
 
 impl FromStr for AllowedAppIds {
-    type Err = HcHttpGatewayError;
+    type Err = Error;
```
cdunster, line 114: This could just be `Infallible` as I don't see any errors being returned.

src/config.rs
```diff
@@ -202,13 +185,13 @@ pub struct ZomeFn {
 }
 
 impl FromStr for AllowedFns {
-    type Err = HcHttpGatewayError;
+    type Err = std::io::Error;
```
cdunster, line 188: I feel like this should be a custom Parse error because there is no IO happening in this function so returning an IO error doesn't really fit.

tests/websocket.rs
jost-s: actually there's no ooint to return a result here at all if it's always unwrapped https://github.com/holochain/hc-http-gw/pull/25/commits/fbb829168a885f27395aadfd1d3f5fe961efa5f3

src/config.rs
jost-s: I agree, this should be a custom error and outside of the HcHttpGwError, because it has nothing to do with gateway responses.

src/config.rs
jost-s: https://github.com/holochain/hc-http-gw/pull/25/commits/a9b5bca3f39573d3bb6fe9728aa287dd2a6b7fd8

src/config.rs
jost-s: https://github.com/holochain/hc-http-gw/pull/25/commits/a9b5bca3f39573d3bb6fe9728aa287dd2a6b7fd8

src/config.rs
ThetaSinner: Nice, I like that logic

ThetaSinner (approved):

ThetaSinner (approved):

PR #26, opened 2025-03-11
Implement app selection logic

ThetaSinner:


src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        assert_eq!(result, Ok(app_info));
+    }
+
+    #[tokio::test]
+    async fn returns_error_if_multiple_apps_match() {
+        let dna_hash = DnaHash::from_raw_32([1; 32].to_vec());
+        let installed_apps = vec![
+            new_fake_app_info("app_1", dna_hash.clone()),
+            new_fake_app_info("app_1", dna_hash.clone()),
```
jost-s, line 253:
```suggestion
            new_fake_app_info("app_2", dna_hash.clone()),
```

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+            &allowed_apps,
+            &admin_websocket,
+        )
+        .await;
+
+        assert_eq!(result, Err(AppSelectionError::MultipleMatching));
+    }
+
+    #[tokio::test]
+    async fn returns_error_if_coordinator_identifier_does_not_match_app_id() {
```
jost-s, line 280: This is practically a duplicate of looking up an app that is not installed. Whether we're currently using coordinator_identifier as app_id shouldn't matter here. I suggest to remove this test.

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+            &allowed_apps,
+            &admin_websocket,
+        )
+        .await;
+
+        assert_eq!(result, Err(AppSelectionError::NotInstalled));
+    }
+
+    #[tokio::test]
+    async fn returns_error_if_matching_coordinator_identifier_not_in_allowed_list() {
```
jost-s, line 308: Same, this is a duplicate of the above.

src/error.rs
```diff
@@ -75,6 +80,18 @@ impl IntoResponse for HcHttpGatewayError {
                 StatusCode::BAD_GATEWAY,
                 Json(ErrorResponse::from("Could not connect to Holochain")),
             ),
+            HcHttpGatewayError::AppSelectionError(AppSelectionError::NotInstalled) => (
```
jost-s, line 83: I find the app selection errors are overrepresented.

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+///
+/// # Side effects
+/// If a matching app is not found in the provided list of installed apps then a request to the
+/// admin websocket will be made and the list will be updated with the results of that request.
+pub async fn try_get_valid_app(
+    dna_hash: DnaHash,
+    coordinator_identifier: String,
+    installed_apps: AppInfoCache,
+    allowed_apps: &AllowedAppIds,
+    admin_call: impl Deref<Target = impl AdminCall + ?Sized>,
```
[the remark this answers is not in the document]
ThetaSinner: I wondered the same, this makes it possible to pass the mock in directly in unit tests which is rather neat

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+            } else {
+                // We either couldn't get a response from Holochain or the response was empty.
+                // In either case, we can't find the app.
+
+                return Err(AppSelectionError::NotInstalled);
+            }
+        }
+    };
+
+    if !allowed_apps.contains(&app_info.installed_app_id) {
```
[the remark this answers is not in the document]
ThetaSinner: I'd like to avoid using the `coordinator_identifier` as an app id anywhere except in the app lookup logic. It's going to be a pain to unpick otherwise. As soon as we've done the app lookup we know the app id but before that I'd hope we don't need it?

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+        );
+        return Err(AppSelectionError::NotAllowed);
+    }
+
+    Ok(app_info)
+}
+
+fn choose_unique_app<'a>(
+    dna_hash: &DnaHash,
+    coordinator_identifier: &str,
```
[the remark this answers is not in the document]
ThetaSinner: As above, I'd prefer we don't do that

src/app_selection.rs
ThetaSinner: This is tricksy. I thought this looked odd but didn't check why on the first pass of this code. The test breaks if you change this. 

Because we're using the coordinator identifier as an app id, the only way we can test collisions is to have two apps with the same id. That's impossible on Holochain but we can construct test data that way.

I'm okay with having this here as a placeholder but it is thoroughly odd for now :)

src/app_selection.rs
ThetaSinner: Agreed, removed

src/app_selection.rs
ThetaSinner: Agreed, removed

src/app_selection.rs
```diff
@@ -0,0 +1,404 @@
...
+use holochain_conductor_api::CellInfo;
+use holochain_types::dna::DnaHash;
+use std::ops::Deref;
+use std::sync::Arc;
+use thiserror::Error;
+
+use crate::{config::AllowedAppIds, AdminCall};
+
+#[derive(Debug, PartialEq, Error)]
+pub enum AppSelectionError {
```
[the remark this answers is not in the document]
ThetaSinner: I'm a little on the fence with this one. We are returning different HTTP statuses depending on what happened and the errors are checked in the unit tests. I think this is meeting the threshold where distinguishing errors is useful. I'm trying to think how this could be done cleanly with strings and I think everything except the HTTP status codes could.

I know you cleaned up the error handling in general earlier and maybe I'm too tired to see what needs doing here?

src/error.rs
ThetaSinner: Responded above, I'm not sure I've got a clear picture of what to do about this one

src/app_selection.rs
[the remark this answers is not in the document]
jost-s: It seems this is an answer to a different comment.

src/error.rs
jost-s: Okay, if you want different status codes, then let's keep this.

jost-s (approved):

src/app_selection.rs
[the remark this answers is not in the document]
ThetaSinner: I'm hoping it's just in one place. If we treat it as opaque except in the app selection logic, that should be the only place we have to touch it. Maybe validation also needs tweaking to deserialize it but beyond that it should be minimal 🤞🏻

src/app_selection.rs
```diff
@@ -0,0 +1,353 @@
...
+        let installed_apps = installed_apps.read().await;
+        choose_unique_app(&dna_hash, &coordinator_identifier, &installed_apps)
+            .ok()
+            .cloned()
+    };
+
+    let app_info = match app_info {
+        Some(app_info) => app_info,
+        None => {
+            let mut installed_apps = installed_apps.write().await;
```
mattyg, line 53: Should we move this write lock down to where it is actually used, rather than before waiting for a admin ws call?

ThetaSinner: In this case yes, it may as well. We aren't actually preventing repeated lookups here so we may as well be light on the lock

I'll do this in a follow up

src/app_selection.rs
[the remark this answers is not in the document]
jost-s: It's here in app selection, in the zome call route where it's checked whether a function is allowed by coordinator_identifier  and in the UnauthorizedFunction error where the identifier is standing in as app id.

src/app_selection.rs
cdunster: Good that the test worked then :grinning:

src/app_selection.rs
cdunster: Although the outcome should be the same, this was actually a distinct test because the other test is testing the scenario where the DNA hash can't be found and this test was testing the scenario where it could be found, but the `app_id` was different. So now, if we remove the `app_id` check from the code, then all the tests still pass, meaning we have a bug. I know that checking the `app_id` should be tempory but I'd still like to have a test for it :slightly_smiling_face:

src/app_selection.rs
cdunster: Similar to my comment above, this tested if two installed apps contain the same DNA hash, but the allowed app does not have the expected `app_id`.

src/app_selection.rs
ThetaSinner: See https://github.com/holochain/hc-http-gw/pull/31

PR #27, opened 2025-03-11
Thetasinner feat/establish admin ws connection

ThetaSinner:


src/config.rs
```diff
@@ -22,9 +22,6 @@ pub const DEFAULT_ZOME_CALL_TIMEOUT: std::time::Duration = std::time::Duration::
 /// Errors when parsing config arguments.
 #[derive(Debug, thiserror::Error)]
 pub enum ConfigParseError {
-    /// Error when parsing URL.
-    #[error("URL parse error: {0}")]
-    UrlParseError(#[from] url2::Url2Error),
```
ThetaSinner, line 27: This was such a pain to work with. Within the lib it is now only a `SocketAddr`. That makes setting up app connections to the same IP really simple and avoids having to extract parts of the URL in multiple places.

src/holochain.rs
```diff
@@ -43,10 +43,6 @@ pub trait AdminCall: std::fmt::Debug + Send + Sync {
         allowed_origins: AllowedOrigins,
         installed_app_id: Option<String>,
     ) -> BoxFuture<'static, HcHttpGatewayResult<u16>>;
-
-    /// Set the admin websocket connection to use.
-    #[cfg(feature = "test-utils")]
```
ThetaSinner, line 48: No longer required, the admin socket reconnects and we don't need to simulate it any more

src/holochain/admin_conn.rs
```diff
@@ -1,47 +1,154 @@
...
     fn issue_app_auth_token(
         &self,
-        _payload: IssueAppAuthenticationTokenPayload,
+        payload: IssueAppAuthenticationTokenPayload,
     ) -> BoxFuture<'static, HcHttpGatewayResult<AppAuthenticationTokenIssued>> {
-        todo!()
+        let this = self.clone();
+        Box::pin(async move {
+            this.call(|admin_ws| {
+                let payload = IssueAppAuthenticationTokenPayload {
```
ThetaSinner, line 106: We should probably make this clone, but that doesn't justify doing a release of 0.4 so for now this will do. A change we should make on 0.5 I think

src/holochain/app_conn_pool.rs
```diff
@@ -184,21 +184,11 @@ impl AppConnPool {
...
-
-        tracing::info!("Connecting to app websocket at {}:{}", host, app_port);
-
         // Build a connection request
-        let request = ConnectRequest::from(
-            format!("{host}:{app_port}")
-                .parse::<SocketAddr>()
-                .expect("Host and app port must be valid"),
-        )
+        let request = ConnectRequest::from(SocketAddr::new(
```
ThetaSinner, line 188: So much nicer

src/resolve.rs
```diff
@@ -0,0 +1,49 @@
...
+    let url = Url::parse(url).map_err(|e| std::io::Error::other(format!("Invalid URL: {}", e)))?;
+
+    let host = url
+        .host_str()
+        .ok_or_else(|| std::io::Error::other("Missing host"))?;
+    let port = url
+        .port()
+        .ok_or_else(|| std::io::Error::other("Missing port"))?;
+
+    let maybe_addr = lookup_host(format!("{host}:{port}")).await?.next();
```
ThetaSinner, line 15: This is what `ToSocketAddrs` calls. So if somebody uses `ws://localhost:8888` rather than `ws://127.0.0.1:8888`, we won't crash and burn

src/router.rs
```diff
@@ -68,27 +70,37 @@ pub mod tests {
...
                     };
-                    Ok(ExternIO::encode(response)
-                        .map_err(|e| HcHttpGatewayError::RequestMalformed(e.to_string()))?)
+                    ExternIO::encode(response)
+                        .map_err(|e| HcHttpGatewayError::RequestMalformed(e.to_string()))
                 })
             });
 
+            // TODO This should be a mock. Done on another PR...
+            let admin_call = AdminConn::new(config.admin_socket_addr);
```
ThetaSinner, line 99: For now, because app lookup and zome calls aren't implemented, this isn't getting used. So it doesn't do any harm. Will disappear on a rebase.

tests/websocket.rs
```diff
@@ -553,71 +545,3 @@ fn create_test_configuration(admin_port: u16) -> Configuration {
     )
     .unwrap()
 }
-
-impl AdminCall for AdminWrapper {
```
ThetaSinner, line 557: No longer needed, we can use the real one

tests/zome_call.rs
```diff
@@ -47,16 +48,28 @@ async fn zome_call_with_valid_params_but_no_payload() {
...
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
 async fn zome_call_with_payload_exceeding_limit_fails() {
     initialize_tracing_subscriber();
 
     let mut allowed_fns = std::collections::HashMap::new();
     allowed_fns.insert("forum".to_string(), AllowedFns::All);
 
+    let sweet_conductor = SweetConductor::from_standard_config().await;
```
ThetaSinner, line 58: Again, wrong, will be fixed on a rebase

src/app_selection.rs
```diff
@@ -51,10 +51,13 @@ pub async fn try_get_valid_app(
         Some(app_info) => app_info,
         None => {
             let mut installed_apps = installed_apps.write().await;
-            let new_installed_apps = admin_call.list_apps().await.unwrap_or_else(|e| {
-                tracing::error!("Failed to get a list of apps from Holochain: {}", e);
-                vec![]
-            });
+            let new_installed_apps = admin_call
+                .list_apps(Some(AppStatusFilter::Running))
```
ThetaSinner, line 55: This had been missed without a real admin websocket. We do actually want to pass a filter because there's no point connecting to an app that isn't running.

src/holochain/admin_conn.rs
```diff
@@ -1,51 +1,178 @@
...
+                return Ok(admin_ws.clone());
+            }
+        }
+
+        let mut lock = self.handle.write().await;
+
+        // We might have been queued up behind another task that was holding the write lock, so we
+        // need to check again after obtaining the write lock. Reconnecting if another task has
+        // already reconnected risks closing the connection the other task just established.
+        if let Some(admin_ws) = lock.as_ref() {
```
ThetaSinner, line 79: Only just noticed this. We need to check this again after obtaining a write lock. I've put the same change into the app websocket logic

jost-s (approved):
Hooray! Just a suggestion

src/holochain/admin_conn.rs
```diff
@@ -1,51 +1,178 @@
...
+    /// to the specified socket address.
+    pub fn new(socket_addr: SocketAddr) -> Self {
+        Self {
+            socket_addr,
+            handle: Default::default(),
+        }
+    }
+
+    /// Allows calling a method on the [`AdminWebsocket`], with automatic reconnection if needed
+    pub async fn call<T>(
```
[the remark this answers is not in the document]
ThetaSinner: Yep, good spot

jost-s (approved):

PR #28, opened 2025-03-12
chore: Split testing and runtime logging

ThetaSinner:


src/lib.rs
```diff
@@ -9,9 +9,11 @@ mod holochain;
 mod router;
 mod routes;
 mod service;
-pub mod tracing;
 pub mod transcode;
 
+#[cfg(any(test, feature = "test-utils"))]
+pub mod tracing;
```
ThetaSinner, line 15: Notice, now only available for testing. The library doesn't need to be deciding how people log when consuming it.

jost-s (approved):

src/lib.rs
ThetaSinner: Renamed the module and function, and updated the module doc comment

src/tracing.rs
```diff
@@ -1,22 +1,20 @@
...
+    fmt::{self, time::UtcTime},
     layer::SubscriberExt,
     EnvFilter, Registry,
 };
 
-const DEFAULT_LEVEL: &str = "info";
+const DEFAULT_LOG_LEVEL: &str = "warn,holochain_http_gateway=info";
 
 /// Initialize a global tracing subscriber
 pub fn initialize_tracing_subscriber() {
```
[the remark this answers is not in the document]
ThetaSinner: Updated the module name, function name, and added some details to the doc comment

c12i (approved):

PR #29, opened 2025-03-12
Hook up zome call

jost-s:
Closes the loop for making a zome call. Adds a unit test for the route.


src/app_selection.rs
```diff
@@ -128,7 +128,7 @@ mod tests {
     use std::str::FromStr;
     use tokio::sync::RwLock;
 
-    fn new_fake_app_info(app_id: impl ToString, dna_hash: DnaHash) -> AppInfo {
+    pub fn new_fake_app_info(app_id: impl ToString, dna_hash: DnaHash) -> AppInfo {
```
jost-s, line 131: Exposing this to be re-used. Could be moved to some shared test module.

src/routes/zome_call.rs
```diff
@@ -122,182 +125,42 @@ pub async fn zome_call(
...
-        ExternIO::encode(()).map_err(|err| {
-            HcHttpGatewayError::RequestMalformed(format!("Failure to serialize payload - {err}"))
-        })?
-    };
-
-    Ok(())
-}
-
-#[cfg(test)]
-mod tests {
```
jost-s, line 138: Moved to `validations.rs`

src/routes/zome_call.rs
```diff
@@ -122,182 +125,42 @@ pub async fn zome_call(
...
+            CellInfo::Provisioned(provisioned_cell) => {
+                if *provisioned_cell.cell_id.dna_hash() == dna_hash {
+                    Some(provisioned_cell.cell_id.clone())
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        })
+        // The app info has been found based on the DNA hash, so the cell must exist
```
jost-s, line 146: Really messed up diff

src/transcode.rs
```diff
@@ -8,16 +8,26 @@
...
-pub fn base64_json_to_hsb(base64_encoded_payload: &str) -> HcHttpGatewayResult<ExternIO> {
-    let base64_decoded_payload = BASE64_URL_SAFE
-        .decode(base64_encoded_payload)
-        .map_err(|_| HcHttpGatewayError::RequestMalformed("Invalid base64 encoding".to_string()))?;
-    let json_payload = serde_json::from_slice::<serde_json::Value>(&base64_decoded_payload)
-        .map_err(|_| HcHttpGatewayError::RequestMalformed("Invalid JSON value".to_string()))?;
+/// Function to transcode an optional base64 encoded payload to Holochain serialized bytes
+/// (type `ExternIO`). If no payload is passed in, a unit value will be serialized.
+pub fn base64_json_to_hsb(
+    maybe_base64_encoded_payload: Option<String>,
```
jost-s, line 17: Accepting the optional payload here. That makes it cleaner in the calling fn.

src/app_selection.rs
ThetaSinner: Agree, we have a few things we should do that with, same with the app info creation

src/router.rs
```diff
@@ -91,53 +82,28 @@ pub mod tests {
...
-                        .map_err(|e| HcHttpGatewayError::RequestMalformed(e.to_string()))
-                })
-            });
+            app_call
+                .expect_handle_zome_call()
+                .returning(|_, _, _, _, _| {
+                    Box::pin(async move { Ok(ExternIO::encode(()).unwrap()) })
+                });
+            let app_call = Arc::new(app_call);
+            Self(hc_http_gateway_router(config, admin_call, app_call))
```
ThetaSinner, line 98: `Self::new_with_config_and_interfaces`?

src/routes/zome_call/tests/responses.rs
```diff
@@ -0,0 +1,54 @@
...
+use crate::{MockAdminCall, MockAppCall};
+use holochain_client::ExternIO;
+use holochain_types::prelude::DnaHash;
+use reqwest::StatusCode;
+use std::collections::HashMap;
+use std::net::{Ipv4Addr, SocketAddr};
+use std::sync::Arc;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn happy_zome_call() {
```
ThetaSinner, line 15: Nice!

src/transcode.rs
```diff
@@ -8,16 +8,26 @@
...
+        let base64_decoded_payload =
+            BASE64_URL_SAFE
+                .decode(base64_encoded_payload)
+                .map_err(|_| {
+                    HcHttpGatewayError::RequestMalformed("Invalid base64 encoding".to_string())
+                })?;
+        serde_json::from_slice::<serde_json::Value>(&base64_decoded_payload)
+            .map_err(|_| HcHttpGatewayError::RequestMalformed("Invalid JSON value".to_string()))?
+    } else {
+        serde_json::Value::Null
```
ThetaSinner, line 29: Ahh, nice okay. I was wondering about this last night. It seems silly for people to have to encode null and pass it if they want to call a zome function that doesn't have parameters. This isn't in the spec but we should just add it.

ThetaSinner (approved):

src/router.rs
ThetaSinner: Done: https://github.com/holochain/hc-http-gw/pull/29/commits/a3fe0561bda540ab743d2befe9d3c14c081be888

ThetaSinner (approved):

src/holochain/app_conn_pool.rs
```diff
@@ -328,9 +328,35 @@ impl AppCall for AppConnPool {
...
+                    let result = app_ws
+                        .call_zome(
+                            ZomeCallTarget::CellId(cell_id),
+                            zome_name.into(),
+                            fn_name.into(),
+                            payload,
+                        )
+                        .await;
+                    if let Err(err) = &result {
+                        tracing::debug!(?err);
```
c12i, line 353: Shouldn't this be an error log?

ThetaSinner: There could potentially be many of these and the administrator can't do much about it if the app doesn't have the data that's being requested or is coded wrong. At most, we would monitor the gateway for an error response count and come looking at what's going on if that spikes

PR #31, opened 2025-03-12
Clean up fixes

ThetaSinner:


src/app_selection.rs
```diff
@@ -63,6 +62,7 @@ pub async fn try_get_valid_app(
                 // If we got a response from Holochain, then we have a chance of finding the app.
                 // Update the app info cache and search again.
 
+                let mut installed_apps = installed_apps.write().await;
                 *installed_apps = new_installed_apps.clone();
                 choose_unique_app(&dna_hash, &coordinator_identifier, &installed_apps)?.clone()
```
cdunster, line 67: This write lock could be downgraded here into a read lock. I don't think that it makes a difference right now as this is the only place that it's used, and a write lock is always required first, but maybe in the future.

```suggestion
                choose_unique_app(
                    &dna_hash,
                    &coordinator_identifier,
                    &installed_apps.downgrade(),
                )?
                .clone()
```

src/config.rs
```diff
@@ -14,7 +14,7 @@ use std::{
 pub const DEFAULT_PAYLOAD_LIMIT_BYTES: u32 = 10 * 1024;
 
 /// Default maximum number of app connections that the gateway will maintain concurrently.
-pub const DEFAULT_MAX_APP_CONNECTIONS: u32 = 500;
+pub const DEFAULT_MAX_APP_CONNECTIONS: u32 = 50;
```
cdunster, line 17: Is this value a requirement anywhere or just a gut feeling?

ThetaSinner: The second option. I doubt we can handle 500 apps at once and if somebody wants to try I'd prefer they have to configure it. I was thinking of taking this down to something more like 20

src/app_selection.rs
ThetaSinner: That's a good point. I was thinking the lock was not going to be held much longer because the choose is quick, but it's an easy fix for a little less time holding the lock

Cargo.toml
```diff
@@ -37,10 +38,17 @@ url = "2.5"
...
+  "sqlite-encrypted",
+  "tx5",
+  "wasmer_sys",
+  "sweettest",
+] }
 http-body-util = "0.1.2"
 mockall = "0.13"
-reqwest = { version = "0.12", features = ["rustls-tls"] }
+reqwest = { version = "0.12", default-features = false, features = [
+  "rustls-tls",
```
jost-s, line 50: It looks like we're not using TLS anywhere in the tests. Can be dropped too?

Cargo.toml
```diff
@@ -37,10 +38,17 @@ url = "2.5"
 holochain_http_gateway = { path = ".", features = ["test-utils"] }
 
 assert2 = "0.3.15"
-holochain = { version = "0.4.1", features = ["sweettest"] }
+holochain = { version = "0.4.1", default-features = false, features = [
```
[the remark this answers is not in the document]
jost-s: Is the `json` feature of the tracing subscriber still used?

jost-s (approved):
Funtastic

src/test.rs
```diff
@@ -0,0 +1,6 @@
+//! Test support module
+
+pub mod data;
+#[cfg(test)]
```
cdunster, line 4: Is the idea that the `test` module can be used by other crates to test things using the `test-utils` feature but the `test::router` module can't be, is that necessary?

Cargo.toml
ThetaSinner: Oh yeah, du'h :) Fixed

Cargo.toml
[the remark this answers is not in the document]
ThetaSinner: Nope, tidied!

PR #33, opened 2025-03-12
feat(error): differentiate zome call error responses

jost-s:
Test individual error responses from a zome call.


src/holochain/app_conn_pool.rs
```diff
@@ -334,23 +334,32 @@ impl AppCall for AppConnPool {
         payload: ExternIO,
     ) -> BoxFuture<'static, HcHttpGatewayResult<ExternIO>> {
         let this = self.clone();
+        let app_id = installed_app_id.clone();
```
ThetaSinner, line 337: Clone twice?

src/routes/zome_call/tests/responses.rs
```diff
@@ -1,54 +1,206 @@
...
+async fn happy_zome_call() {
     let mut app_call = MockAppCall::new();
     app_call
         .expect_handle_zome_call()
         .returning(|_, _, _, _, _| {
             Box::pin(async move { Ok(ExternIO::encode("return_value").unwrap()) })
         });
-    let app_call = Arc::new(app_call);
-    let router = TestRouter::new_with_config_and_interfaces(config, admin_call, app_call);
+    // let app_call = Arc::new(app_call);
```
ThetaSinner, line 51: Commented?

src/routes/zome_call/tests/responses.rs
```diff
@@ -1,54 +1,206 @@
...
+                ))
+            })
+        });
+    let router = create_test_router(app_call);
+    let (status_code, body) = router
+        .request(&format!("/{DNA_HASH}/{APP_ID}/coordinator/fn_name"))
+        .await;
+    assert_eq!(status_code, StatusCode::INTERNAL_SERVER_ERROR);
+    assert_eq!(body, r#"{"error":"Something went wrong"}"#);
+}
```
ThetaSinner, line 206: These tests look correct for these error types. How about errors from Holochain that indicate a failure within the zome call that should go back to the user? If possible we want to make this debugable by the end user.

"JSON error response with an error field with a hard-coded string for conductor errors or the zome error message if this was an error raised by the target hApp."

src/holochain/app_conn_pool.rs
jost-s: Yup, every one of these is required.

jost-s: Yes, every one of these is required.

src/routes/zome_call/tests/responses.rs
jost-s: Yes, all we have here is what the AppWebsocket returns, and that's a `holochain_client::ConductorApiError`. That has to be modified then.

ThetaSinner: Right, right, we talked about this and it's all strings for now. Okay fine with me. 

We might find that we get questions about this and we need to look at upping the log level for failed zomes so there's a way in for debugging.

src/error.rs
```diff
@@ -51,52 +50,38 @@ impl From<String> for ErrorResponse {
...
+                "Something went wrong".to_string(),
             ),
         }
-        .into_response()
+    }
+}
+
+impl IntoResponse for HcHttpGatewayError {
+    fn into_response(self) -> axum::response::Response {
+        let (status_code, body) = self.into_status_code_and_body();
```
jost-s, line 84: I've been thinking since I first saw this module that there's got to be a way to make this less verbose. Happy now.

.github/workflows/test.yaml
```diff
@@ -15,18 +15,18 @@ jobs:
...
-        with:
-          name: holochain-ci
       - name: Maximize build space
         uses: AdityaGarg8/remove-unwanted-software@v5
         with:
           remove-dotnet: 'true'
           remove-android: 'true'
           remove-codeql: 'true'
           remove-docker-images: 'true'
+      - uses: Swatinem/rust-cache@v2
```
jost-s, line 25: Cache couldn't be restored without maximizing space first.

ThetaSinner (dismissed):

ThetaSinner (approved):

Issue #42, opened 2025-03-13
Update to Holochain 0.5


ThetaSinner:
The initial version of the gateway runs against 0.4. As we do with other projects, we should keep the gateway updated and working with the latest dev version of Holochain for people who want to develop against the latest version of Holochain. It also helps us maintain awareness of the impact changes in Holochain are having downstream.

PR #48, opened 2025-07-04
ci: add cron workflows to update flake and cargo lockfiles

cdunster:
As part of #42.


.github/workflows/cargo_update.yaml
```diff
@@ -0,0 +1,32 @@
...
+        uses: cachix/install-nix-action@v31
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.28.3/install
+
+      - name: Flake update
+        run: |
+          nix develop .#ci --command cargo update
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
```
ThetaSinner, line 25: Should this use a token from the secrets that should be available on most repos now for release automation? Then the workflows will run because the change it submitted by a trusted user. I didn't want to maintain tokens by hand across many repos which is why this workflow is set up in other places to use the default token to create the PR.

ThetaSinner (approved):
Change looks good. A suggestion for possible improvement to how many clicks we need for these PRs :)

jost-s (approved):

PR #49, opened 2025-07-04
chore: update `serde_json` to use Holochain fork

cdunster:
As part of #42.

The fork contains binary array deserialization.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ holochain_conductor_api = "0.4.1"
 holochain_types = "0.4.1"
 holochain_websocket = "0.4.1"
 serde = { version = "=1.0.203", features = ["derive"] }
-serde_json = { version = "1.0.140", git = "https://github.com/ThetaSinner/json.git", branch = "master" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ holochain_conductor_api = "0.4.1"
 holochain_types = "0.4.1"
 holochain_websocket = "0.4.1"
 serde = { version = "=1.0.203", features = ["derive"] }
-serde_json = { version = "1.0.140", git = "https://github.com/ThetaSinner/json.git", branch = "master" }
+serde_json = { package = "hc_serde_json", version = "1.0.141" }
```
cdunster, line 26: I could also add it as a patch so that all `serde_json` usages from crates we depend on also use the fork, but I don't think there will be any compatibility issues with just adding it to this crate.

jost-s: In Holochain only the deserialization of the zome call command in the sandbox is patched. But here serde_json is only used in one place anyway, so I think this works well.

jost-s (approved):

PR #50, opened 2025-07-04
chore: backport CI workflows to 0.4 branch

cdunster:
[Hoisted by my own petard](https://github.com/holochain/hc-github-config/pull/45#discussion_r2171610697) I need a review to update the `main-0.4` branch, that I only just created, as I think that we want all of these changes in the `0.4` release before upgrading to holochain `0.5`.


ThetaSinner (approved):

ThetaSinner:
Interesting. We do have a team agreement to not need a review on backports so even if we don't have admin access, perhaps we need to tweak the settings for review requirements on release branches?

PR #55, opened 2025-07-08
chore: migrate to holochain v0.5

### LINKED ISSUE #42: Update to Holochain 0.5

cdunster:
Update the gateway to use holochain v0.5 instead of v0.4.

Closes #42.


Cargo.toml
```diff
@@ -18,11 +18,11 @@ axum = "0.8.1"
...
-holochain_client = "0.6.2"
-holochain_conductor_api = "0.4.1"
-holochain_types = "0.4.1"
-holochain_websocket = "0.4.1"
-serde = { version = "=1.0.203", features = ["derive"] }
+holochain_client = "0.7.1"
+holochain_conductor_api = "0.5.3"
+holochain_types = "0.5.3"
+holochain_websocket = "0.5.3"
+serde = { version = "=1.0.219", features = ["derive"] }
```
ThetaSinner, line 25: I think we don't have to pin serde if we depend on `holochain_serialized_bytes` because that pins it for us. It doesn't hurt but could maybe save some future typing by setting this to "1"?

ThetaSinner (dismissed):
Nice!

ThetaSinner (approved):

PR #56, opened 2025-07-09
chore: add missing CI devShell in flake.nix

cdunster:
The [`cargo_update` CI workflow](https://github.com/holochain/hc-http-gw/blob/da9a0524ef1eb5ae2bc42d0362b5b01949e74c17/.github/workflows/cargo_update.yaml#L21) uses a devShell `ci` to perform the `cargo update` command, and that devShell doesn't exist right now.

The `ci` devShell is only used to perform `cargo update`, so none of the build or test dependencies should be required.


jost-s (approved):
👍

PR #57, opened 2025-07-09
Prepare next release

holochain-release-automation2:
Automated preparation for the next release, triggered by cdunster.


ThetaSinner (changes requested):
Don't we need to release 0.2.0 for the version that works with 0.5?

cdunster:
> Don't we need to release 0.2.0 for the version that works with 0.5?

Yeah, I changed it locally. I was just interested in seeing if the automated release would pick up on the fact that the minor version needed bumping because the minor versions of 0-release dependencies got bumped.

jost-s (approved):

ThetaSinner:
Would be cool if it was that smart :D It'd either need to understand our branching model or maybe notice incompatible dependencies from Holochain or something.

PR #103, opened 2026-07-20
docs: update local gateway workflow

### LINKED ISSUE #93: Update hc-sandbox usage in the README
I tried to test the gateway with the latest version of Holochain and found that the [instructions in the README](https://github.com/holochain/hc-http-gw#instructions) seem to be out of date. Specifically the `hc sandbox call ...` command is no longer a valid command.

We should update the instructions and make sure that the gateway still works okay.

It should be a matter of adding the `holochain_cli_client` binary and changing the commands in the README to `hc client call ...`.

AC:

- [x] Verify that the gateway still runs with the test fixture
- [x] Update the README with instructions on how to run the gateway locally with the test fixture

veeso:
## What changed

The local setup guide now uses the current hc client commands for installing the fixture, listing DNA hashes, authorizing zome calls, and creating fixture data. It also explains how to carry the conductor admin port between terminals and records the Holochain 0.7.x compatibility range.

## Why

The previous guide used hc sandbox commands that are no longer available, so a new checkout could not follow the documented workflow.

## Notes

Verified end to end in the Nix development shell. Fixture packaging, gateway health checks, read and write calls, the full Rust test suite, and Markdown lint all pass.

Closes #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 [![API Docs](https://docs.rs/holochain_http_gateway/badge.svg)](https://docs.rs/holochain_http_gateway)
 [![Discord](https://img.shields.io/badge/Discord-blue.svg?style=flat-square)](https://discord.gg/k55DS5dmPH)
 
-The Holochain HTTP Gateway for providing a way to bridge from the web2 world into Holochain. Read the [spec](./spec.md) for more details.
+The Holochain HTTP Gateway provides a way to bridge from the web2 world into
+Holochain. Read the [spec](./spec.md) for more details.
 
 ## Compatibility
 
@@ -13,13 +14,21 @@ The Holochain HTTP Gateway for providing a way to bridge from the web2 world int
 | 0.4.x             | 0.1.x                |
 | 0.5.x             | 0.2.x                |
 | 0.6.x             | 0.3.x                |
+| 0.7.x             | 0.4.x                |
 
 ## Running HTTP Gateway locally
 
 ### Prerequisites
 
-Enter the Nix `devShell` with `nix develop` or make sure that you have `rustup`
-and `holochain-cli` installed.
+Enter the Nix development shell. It provides the versions of Rust, Holochain,
+and the `hc` CLI that are compatible with this checkout:
+
+```bash
+nix develop
+```
+
+Run the commands below from the repository root. Open each new terminal with
+`nix develop` before continuing in it.
 
 ### Instructions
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,52 +39,62 @@ hc sandbox clean
 ```
 
 Create a new sandboxed conductor with the lair keystore running in the
-Holochain process, entering a passphrase when prompted:
```
ThetaSinner, line 33: I just ran `hc s create` at 0.7.0-rc.1 and it does still prompt for a passphrase at this point. It shouldn't just be at the `run` subcommand, because the setup should involve creating and encrypting the keystore. Did something else happen while testing for this change?

README.md
```diff
@@ -30,52 +39,62 @@ hc sandbox clean
...
-first with the `package` script:
+In a second terminal, enter the Nix development shell and set `ADMIN_PORT` to
+the value printed by `hc sandbox run`:
+
+```bash
+nix develop
+export ADMIN_PORT=41191
+```
+
+Build and package the test fixture (`fixture1`):
```
ThetaSinner: The text that was here about using any hApp you want seemed reasonable. I guess it's okay that it's just implied that you can swap this out with your own hApp?

README.md
```diff
@@ -30,52 +39,62 @@ hc sandbox clean
...
 
-In a separate terminal, also in the Nix `devShell`, we need to install the hApps that you want to use.
-For example, we can use the test fixture (`fixture1`) but it needs to be built
-first with the `package` script:
+In a second terminal, enter the Nix development shell and set `ADMIN_PORT` to
+the value printed by `hc sandbox run`:
+
+```bash
+nix develop
+export ADMIN_PORT=41191
```
cdunster, line 67: I'd prefer it if all of these code blocks were a single command so that someone can easily copy-paste them from GitHub. Not a big deal, though.

README.md
```diff
@@ -30,52 +39,62 @@ hc sandbox clean
...
 ```
 
-In another new terminal, also in the Nix `devShell`, check that we get a response to the health-check:
+In a third terminal, enter the Nix development shell and check the health
+endpoint:
 
 ```bash
+nix develop
 curl -i localhost:8090/health
 ```
```
cdunster, line 100: This is a good example; I'd prefer either:

~~~suggestion
In a third terminal, enter the Nix development shell

```bash
nix develop
```

Then, check the health endpoint:

```bash
curl -i localhost:8090/health
```
~~~

Alternatively:


~~~suggestion
In a third terminal, enter the Nix development shell and check the health
endpoint:

```bash
nix develop -c curl -i localhost:8090/health
```
~~~

cdunster (dismissed):
Some suggestions but nothing blocking.

README.md
veeso: sorry, I didn't notice codex randomly removed this phrase with no reason. It still prompts.

README.md
veeso: You still couldn't because the port is random everytime and you need to manually check it from the terminal

README.md
veeso: applied

README.md
veeso: restored

cdunster (approved):

